PR #257, opened 2026-06-11
fix: dogfood MED/LOW batch 2 + close the C3 fork-bomb for ALL --json+passthrough combos

oimiragieo:
## Summary

Final batch of the dogfood MEDIUM/LOW sweep — the cross-file/judgment-call items, handled conservatively (additive on contracts, assess-don't-force on the by-design candidates).

| ID | Outcome |
|---|---|
| **M1** | `tg --json` and `--json --stats` emit the same match schema — JsonFormatter now carries both `line` and `line_number` (mirrors NdjsonFormatter); consumers keyed on `line` survive `--stats` |
| **M3** | every `tg run --json` shape carries consistent `version`/`schema_version`/`mode`/`total_matches` (additive) |
| **M4** | `--batch-rewrite` error names the required shape (rust validator + Python doc) instead of cryptic `$` |
| **M14b** | inline-flag regex (`start(?s).*end`) transparently retries with PCRE2 + stderr note instead of erroring; `--engine pcre2` now honored |
| **L3** | `tg defs` gains additive `class`+`score` fields and a `--class NAME` filter to disambiguate common names |
| **L4** | **assessed by-design** — `impact` is a fast planning signal pointing to `blast-radius`; wiring its caller_tree/score in would duplicate blast-radius and double cost. Redirect reason made actionable instead |

### Caught in review (not trusting "done")
- M3 was incomplete for the `tg run --pattern --lang` path — but that routes to *search*, whose aggregate `--json` contract (native-served) has never carried `schema_version`; scoped M3 to the documented `run_command` schemas and noted the search-contract follow-up.
- M1's real fix lived in `json_fmt.py` (not main.py) — wired it where the agent flagged.

## Test plan
- 44 new/updated regression tests; intended-additive snapshot/assertion updates (M1 `line`, M3 fields, L4 reason).
- Gate green: `ruff`, `ruff format --preview`, `mypy --strict`, `cargo fmt`/`clippy -D warnings`/`test`, `pytest`. The one local failure is the native-disabled timing env-artifact (`test_public_help_falls_back_to_native...`), green on CI where the native binary is present.


oimiragieo:
Added a critical follow-up commit (`87bc7f2`): the C3 fork-bomb was only closed for **render** flags in v1.13.37 — `--json --debug` / `--json --stats` still fork-bombed the native↔python re-exec loop (this is what hung a user's `tg update` self-test and locked tg.exe). Fixed with a `TG_REEXEC_GUARD` env marker (native sets it on spawn; python refuses to delegate search back). Verified end-to-end: zero spawned children, was exit-124 hang. Needs both the native binary and python package, which ship together in this release.

PR #302, opened 2026-06-29
fix(gpu): close 3 parked audit items — cuDF device-bind, installer SHA, GPU Phase-0

oimiragieo:
**Draft** — the assembled, audited-clean result of the full research→council→fan-out→audit→fix→re-audit pipeline for the 3 parked audit items. Not for auto-merge; for review.

## What this closes
- **cuDF GPU device-binding (HIGH):** `_search_uncapped` now binds the selected CUDA device via a `_device_context` scoped to the **in-process** allocation paths only (single-process zero-copy + chunked loop). The two `_search_distributed` fork calls run **outside** any device context (avoids the CUDA-fork hazard); `_device_context` degrades to `nullcontext` without cupy (CPU/CI no-op). The distributed path keeps its existing subprocess `CUDA_VISIBLE_DEVICES` pin — the strong multi-GPU guarantee.
- **Installer script integrity (MED):** Windows install.ps1 now downloads the pinned uv **binary** (dual-arch x86_64 + aarch64) and verifies a committed SHA-256 (`scripts/uv_checksums.json`) **fail-closed** before use, instead of running the remote `.ps1` (which lacks checksum verification, uv#13074). Linux/macOS keep the pinned self-verifying installer (uv ≥0.11.0). **Both committed SHAs were independently downloaded + hashed to confirm correctness.**
- **GPU production-readiness Phase 0 (HIGH, CPU-buildable spine):** an **independent** CPU oracle (rg-equivalent: `\n` split, dot/binary skip, latin-1) wired into the existing `correctness_gate`; 4 GPU telemetry fields + `tg doctor` tiers; loud non-promotional fallbacks (now wired end-to-end into the JSON envelope) + supported-semantics taxonomy; promotion-proof audit.

## Process (the convergence loop)
Exa research → Plan v1 → **thinktank council** (caught 3 traps + 8 must-fixes before any code) → Plan v2 → **6-agent worktree fan-out** → my verification (fixed every defect: cuDF 11-failure nullcontext + lint/mypy, installer test, proof-audit contract) → integrate (no conflicts) → **adversarial audit** (found a HIGH CUDA-fork hazard + 2 MED, all test-uncaught) → fix-wave → **re-audit: SHIP, zero findings**.

## Verification
163 changed-area tests pass; ruff `--preview` + mypy clean on the changed sources. +1579/−219 across 15 files.

## Explicitly GPU-hardware-gated (NOT in this PR — operator runs on local WSL 4070/5070)
[...]

PR #303, opened 2026-06-29
docs: capture 2026-06-29 session learnings (pipeline + verification gate, IDF fragility, tg-session currency)

oimiragieo:
**Draft** — captures durable learnings from the 2026-06-29 session into the repo docs + the tensor-grep skill. Verified: each edit adversarially accuracy-checked vs the real `tg --help` + the learnings ledger; 66 doc-governance tests pass.

## In-repo changes (this PR)
- **AGENTS.md** — (1) the full CONVERGENCE PIPELINE in the dogfood workflow (commit-plan-before-fanout → worktree slices → **orchestrator verification gate** → adversarial audit → re-audit-until-zero → draft PR); (2) post-build adversarial audit as a named stage + "no file:line citation → discarded"; (3) the new `worktree-fanout-verification-gate` skill in the build/release skill list; (4) a "weak spots" note on **BM25/IDF ranking fragility** (corpus-IDF blast-radius invisible to the call graph; Task #4); (5) the uv `.ps1`-lacks-checksum supply-chain rule (qualified: lands with PR #302).
- **CLAUDE.md** — `worktree-fanout-verification-gate` added to the build/release skill pointers.
- **tensor-grep skill (SKILL.md + REFERENCE.md)** — documents `tg session` (12 subcommands, **re-verified vs the real binary**: every subcommand except `open`/`list`/`daemon` takes the `session_id` from `tg session open --json` as its first arg) + the top-level `tg blast-radius-render`. GPU-observability docs deliberately deferred until PR #302 ships them.

## Applied outside the repo (NOT in this PR; in ~/.claude)
- New global skill `worktree-fanout-verification-gate` (the orchestrator gate: worktree agents can't self-test → re-run in the real venv, format ALL touched files, scoped-green ≠ CI-green).
- `supply-chain-hardening` skill: uv `.ps1` checksum pattern. `authoring-dynamic-workflows` skill: cross-ref + the worktree-can't-self-test lesson.
- Memory: the novel corpus-IDF ranking-fragility insight + CUDA-fork-safe device-binding contract.


oimiragieo:
## Root cause found — and my earlier hypothesis here was WRONG

I previously recorded a hypothesis on this issue that the flake was a **pipe-buffer deadlock**. That is incorrect, and I am correcting it rather than leaving it to send the next reader at healthy code.

**The production path was never at fault.** `run_validation_command` uses `process_control`'s `controlled_with_output`, which drains the pipes *while* timing out. `rust_core/src/main.rs:10458-10463` already documents why a hand-rolled `spawn` + `wait_timeout` + `wait_with_output` must not replace it, and cites the upstream issue (rust-lang/rust#45572). Research confirms that class is real and documented upstream — and that our code is already immune to it.

**The actual defect is in the test, and it is a discrimination failure.** `run_validation_command_captures_large_stdout_without_deadlock` set:

- `timeout_ms = 15_000` (main.rs:3843)
- `assert!(elapsed < Duration::from_secs(10))` (main.rs:3866)

Those contradict. There is a 5-second window where the command **succeeds**, >1MB is captured, the pipes demonstrably drained — and the test fails anyway on the clock. On a loaded windows runner, spawning PowerShell to write 2.6MB legitimately lands in it.

**Why the clock never guarded the deadlock it names:** a real deadlock exhausts the time limit, `terminate_for_timeout` kills the child, `result.success` is `false`, and the *first* assertion fires. The clock can only ever fire where the property under test **holds**. Oracle Form 7 — a measurement with no discriminating power against the defect it cites.

**Fix (PR #802):** kept, not deleted — it does catch a weaker regression `result.success` cannot see (a drain that works but is pathologically slow). Made the bound **derive** from the timeout (`timeout_ms / 2`) so the two can never contradict again, and raised the timeout to 60s so runner slowness is not read as a code defect.

Recurrence data that triggered this: observed on windows **nightly** first, then windows **stable** — broader than the nightly-only signature originally recorded here.

PR #441, opened 2026-07-07
fix(search): -e multi-pattern + -f pattern-file no longer silently collapse (native path + bootstrap front door)

oimiragieo:
HIGH data-loss fix (free-file audit). `tg search --cpu -e foo -e bar` silently dropped `bar`; `-f file` flooded (empty pattern matched all). TWO parts, because the fix must reach the REAL front door:

**1. main.py (native path):** gather all `-e` patterns + `-f` file lines, combine into one rg-parity OR-alternation before the single-pattern backends run; `effective_config` clears fixed_strings when combined; rg path untouched (reads raw `config.regexp`/`file_patterns`). `-f` reads deferred until confirmed-native (an eager read broke the rg-passthrough contract). rg's real `-f` semantics verified empirically (blank lines NOT skipped, no comment syntax, missing file -> exit 2).

**2. bootstrap.py (the critical follow-up, dogfood-surfaced):** the outer argv fast-path delegated `--cpu`/`--json`/`--gpu` searches to the separately-compiled native `tg` binary WITHOUT excluding `-e`/`-f` — so on any machine where that binary resolves, main.py's fix was DEAD CODE. Dogfood-confirmed the native binary is buggy: `--cpu -e foo -e bar` counts the both-match line TWICE, `-f` over-matches. Now `_can_delegate_to_native_tg_search` excludes `-e`/`-f` (routes them through the in-process combine). Governance test flipped: multi-pattern GPU search now routes to the full CLI, not native.

**Verified:** test_cli_bootstrap + test_multi_pattern_native (18 new e2e + the delegation-exclusion test) pass; the flipped governance test; dogfood confirms both-match-once + rg-parity + single-pattern byte-identical + rg path untouched. ruff/format/mypy clean. main.py-only + bootstrap.py (both free files).


oimiragieo:
Closing to unblock the 8 HIGH audit fixes queued on main.py (orient traceback info-disclosure, -f+-o silent-zero, context exit-2 gate, blast-radius exit-1, etc. — audit 2026-07-08) + 3 on bootstrap.py, all blocked by this one-writer PR. This multi-pattern -e/-f feature has a windows-CI-only golden regression (`--json hello` emits submatches on windows where main is green — root cause in the effective_config restructure, not the -e/-f path) that cannot be reproduced on a dev box with rg on PATH. Re-doing the feature with a windows golden test from the start is safer than fixing blind. Branch fix/multipattern-native preserved. Refs task #68/#69.

PR #541, opened 2026-07-12
fix(index): validate --index capability flags at both front doors, fail-closed (#138)

oimiragieo:
## What / why
External-audit remediation #138 item #1 (the flagship correctness blocker). Explicit `--index` silently dropped unsupported flags (`--hidden`/`--max-depth`/`-l`/`-o`/`--replace`/`--sort`/context/`-w`/`-x`/...) and returned **wrong results with exit 0**; the warm auto-index route had the same gap.

## Fix
`index_flag_violations(args, request)` — exhaustive `let SearchArgs { .. }` destructure with **no `..` rest** (a new field fails compilation until classified). 3 classes: **Honor** (14), **PassthroughSafe** (14), **Refuse** (40). Wired to **REFUSE** on the explicit `--index` door and **REROUTE-to-rg** on the warm auto-index door (subsumes 5 of 9 old ad-hoc checks). Folds in: `exit(1)` on zero matches (was exit 0); `-N` un-numbering via `emit_plain_search_matches_with_line_number`; `--cpu`/`--gpu` contradiction Refuse; `--pcre2`-without-rg stays fail-closed. `color` is value-conditional (never/auto honored, always refused).

## Verification (agent, real rust toolchain 1.96.0)
`cargo check` (default + cuda) clean 0 warnings; `cargo fmt --check` clean; `cargo test --bin tg` 80/80; `test_index` 39/39; routing/search/golden 97/97; native/parity/schema/replace 100/102 (2 pre-existing ignored). Manual dogfood against built `tg.exe` confirms refuse/honor/exit-1/contradiction behavior.

## Gate status
**Opus adversarial gate IN PROGRESS** (index surface) — verifying that every Honor/PassthroughSafe flag is actually honored by the trigram index (not silently wrong). A separate front-door gap (Python-passthrough intercepts `--index`+flags before clap) is tracked as a follow-up task, out of scope here. Drains after v1.64.0 on PyPI + gate clears.

### DIFF
--- a/rust_core/src/main.rs
+++ b/rust_core/src/main.rs
@@ -6626,14 +6626,26 @@ fn run_index_query(
     }
 
     if args.count {
-        let unique_count = unique_line_matches(&matches).len();
-        println!("{unique_count}");
-        // fold-in (a): rg exit-parity. The native CPU (run_native_search_with_optional_rg_
-        // fallback) and multi-pattern (emit_multi_pattern_native_results) engines both already
-        // exit(1) on zero matches; run_index_query never did, so `--index --count` on a
-        // no-match query printed "0" but exited 0 -- indistinguishable from "found nothing but
-        // still succeeded" instead of rg's "no match" signal.
-        if unique_count == 0 {
+        // Audit fix #1 must-fix (Opus adversarial gate on PR #541): emit per-file `path:count`
+        // via the SAME emit_count_search_matches the sibling native aggregate path already uses
+        // (emit_multi_pattern_native_results below), NOT a bare aggregate total. The old
+        // `println!("{unique_count}")` printed a single number (e.g. `3`) while every other count
+        // emitter -- `rg -c`, the native CPU engine's append_count_output_bytes, and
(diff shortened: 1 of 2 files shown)


oimiragieo:
Opus adversarial gate returned FIX-FIRST on the one issue: `count` (Honor) emitted a bare aggregate total instead of per-file `path:count`, diverging from native/`rg -c` on explicit `--index -c` AND silently on warm `-c`. Fixed in 0afe499: the count arm now routes through the existing `emit_count_search_matches` (the same helper the sibling native path uses), so both paths emit rg-compatible per-file counts; exit-1-on-no-match preserved. Re-reviewed (Opus): fix is correct, uses the audited helper, 10 pinned tests updated to the per-file shape, order-independent count-map comparison avoids the rg parallel-walk flaky-order trap. Full rust gate green (test_index 42/42, bin 80/80, routing 81). Everything else in the validator was gate-CONFIRMED sound. Drains after v1.64.1 + #540 publish.

PR #542, opened 2026-07-12
fix(backends): repair AstBackend query API (tree-sitter 0.26) + delete dead LSP tensor/GNN + wrapper-preferred routing (#138)

oimiragieo:
## What / why
External-audit remediation #138 item #5. Removes a dead LSP "tensor cache" + GNN path (`_ast_to_graph`, `_update_ast_tensor`, `tensor_cache` — all write-only / unreachable, grep-evidenced) and corrects `AstBackend.is_available()` from CUDA-gated to **tree-sitter-only**, since `AstBackend.search()` is pure tree-sitter query matching with no torch/CUDA. Drops the now-unused `torch_geometric` from the gpu extras.

## Routing change (the flagged trap)
`is_available()` old: `torch_geometric AND tree_sitter AND torch.cuda.is_available()`. New: `tree_sitter present`. `pipeline.py`/`main.py` call it to choose native `AstBackend` vs the `AstGrepWrapperBackend` sidecar — the old gate meant `AstBackend` was **never** selected on non-GPU boxes despite needing no GPU. New **parity test** asserts available-on-tree-sitter-alone; the fail-closed side (tree-sitter absent → False) is unchanged. `search()`'s unavailable guard now raises `BackendExecutionError` (was a bare `RuntimeError`) per the Backend Fail-Closed Contract.

## Verification (agent, real venv)
4 gates green (ruff / format --preview / mypy 79 / pytest 75+4skip) + heavy regression: `test_cli_modes` 477, mcp suite 361, routing/contract 293 — zero regressions.

## Gate status
**Opus parity gate IN PROGRESS** (backends surface: the is_available routing flip + BackendExecutionError change). Drains after v1.64.0 on PyPI + gate clears. Doc residuals (2 skill files, uv.lock torch_geometric) flagged for a follow-up.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -518,8 +518,8 @@ dependencies = [
 # or CUDA-mismatched torch. The SUPPORTED GPU runtime path is the managed installer
 # (scripts/install.sh / install.ps1), which deliberately installs pinned CUDA 12.8 wheels.
 # `pip install tensor-grep[gpu]` is best-effort experimental, not the supported managed GPU path.
-gpu = ["cudf-cu12", "kvikio-cu12", "torch>=2.0", "torch_geometric>=2.5.0"]
-gpu-win = ["torch>=2.0", "torch_geometric>=2.5.0"]
+gpu = ["cudf-cu12", "kvikio-cu12", "torch>=2.0"]
+gpu-win = ["torch>=2.0"]
 nlp = ["transformers>=5.3.0", "tritonclient[http]"]
 ast = ["tree-sitter>=0.22", "tree-sitter-python", "tree-sitter-javascript", "tree-sitter-typescript", "tree-sitter-rust", "tree-sitter-go", "pygls>=1.3.0"]
 # Local hybrid semantic search dense leg (`tg search --semantic`, roadmap #27, Path B Stage 1):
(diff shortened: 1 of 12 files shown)


oimiragieo:
Opus gate returned FIX-FIRST (deletion safe, but the is_available flip was refuted: native query path broken on tree-sitter 0.26 + native/wrapper DSL divergence -> silent `tg scan`/`tg run` results change). Revised in c79f705:
- (B) Fixed the real bug: `_get_query` migrated via `hasattr(language,'query')` -> `tree_sitter.Query(language, source)`; verified live on 0.26 ((function_definition)/(call) match, invalid node types fail-closed with BackendExecutionError).
- (C) Routing invariant restored: `_select_ast_backend_for_pattern` (both ast_workflows.py + main.py) now prefers the ast-grep wrapper WHENEVER available; native is reached ONLY as the ast-grep-absent fallback. 4 new stash-verified tests; existing wrapper-preference routing tests pass unchanged.
- (D) Real-tree-sitter CI test that fails on the pre-fix `Language.query` API (no FakeLanguage mock masking it); gpu-marked AST test converted to CPU.
- is_available-as-CPU-default + DSL reconciliation split to task #141 (out of scope here).

Re-reviewed (Opus diff-review): routing invariant verified against ORIGINAL behavior -> majority (non-CUDA + ast-grep) unchanged (wrapper both before/after); the only delta is CUDA boxes, where the old path selected native and CRASHED on tree-sitter 0.26 -> now correctly uses the wrapper (a fix, not a regression). `_get_query` migration correct. Full suite 4368 passed / 76 skipped. Retitled refactor->fix so it publishes. Drains last, after #540 + #541.

PR #555, opened 2026-07-12
feat(daemon): serve orient + agent capsules via the warm session daemon (Tier-2 moat) (#108)

oimiragieo:
## Summary

Extends task #94's warm-session-daemon fast path (Tier-1: `context-render`/`edit-plan`/the 5 symbol commands) to `tg orient` and `tg agent` (Tier-2) — the two remaining moat commands that previously always paid the full repo-map scan cost even with a daemon already running.

**Autostart-default verification (build-time check, per the spec):** confirmed on current `main` that `_session_daemon_autostart_enabled()` resolves **default-ON** (`src/tensor_grep/cli/main.py:7860`) — unset, or any value other than an explicit falsy token, routes through the daemon; forced OFF in CI/GITHUB_ACTIONS. Tier-2 inherits this flag; no separate flag was added.

## Design

- `build_orient_capsule` / `build_agent_capsule` are now thin cold-path wrappers that build the repo map once and delegate to new `build_orient_capsule_from_map` / `build_agent_capsule_from_map` — cold and warm share one code path, so parity is by construction rather than a second, driftable implementation. `agent`'s map-based core additionally collapses the cold path's **two** independent repo scans (context-render ranking + call-site-evidence blast-radius) into one shared map — a daemon-moat win beyond just skipping the outer scan.
- **TRAP A** (the audit #107 divergence class): the daemon's cached-map call-site-evidence collector (`_collect_capsule_call_site_evidence_from_map`) has no literal-seed rescue, unlike the cold `build_symbol_blast_radius` wrapper (which retries via `_literal_symbol_seed_files` on a truncated no_match). A no_match on a possibly-truncated map is flagged via a new `daemon_evidence_unreliable` sentinel; the client wrapper (`_maybe_agent_via_running_daemon`) discards the whole response and falls back to cold.
- **TRAP B**: `suggested_scope` is computed in the `build_context_render` WRAPPER (repo_map.py), not `build_context_render_from_map`. `build_agent_capsule_from_map` replicates that exact block against its own map so a warm capsule does not silently drop `suggested_scope` on a truncated scan.
- New `"orient"`/`"agent"` daemon commands (`session_store._serve_session_request_from_payload`), dedicated response-cache key builders (`_orient_response_cache_key`/`_agent_response_cache_key`) keyed on every output-affecting flag (mirrors the 5 symbol commands' anti-bleed discipline, audit #113), added-file-sensitive staleness via a new `_ADDED_FILE_SENSITIVE_COMMANDS` set (a new central file changes orient's ranking; a new call site changes agent's evidence), and fail-open client wrappers (native-only, `--gpu-device-ids` always cold since it shells out to a subprocess, autostart-on-miss) mirroring Tier-1's exact shape.
- Reuses the existing `_SessionResponseCache` (32-entry byte-bounded LRU) — no new cache.
- A shared `repo_map._blast_radius_no_match_is_possibly_truncated` predicate now backs all three call sites that need to agree on "is this no_match trustworthy" (the literal-seed-rescue trigger, TRAP A, and the existing Tier-1 `blast-radius` daemon-fallback gate) instead of three independently-drifting inline copies.

## Latency (measured via the real shipped `tg` binary, bootstrap front door, not CliRunner — on the tensor-grep repo itself)

| Command | Cold | Warm (response-cache hit) | Speedup |
|---|---|---|---|
| `tg orient` | 14.58s | 0.95s | ~15.4x |
| `tg agent` | 21.48s | 1.47s | ~14.6x |

Both warm outputs are **byte-identical** to cold (daemon-only provenance fields `session_id`/`routing_reason`/`serve_cache`/`daemon_response_cache` stripped before comparison), verified via a direct JSON diff.

Note on methodology: the *first* warm-mode call after a fresh daemon spawn still pays close to the full cold cost once (the daemon's own session cache is empty until its first request), since `--max-repo-files`-keyed implicit sessions are per-daemon-process, not persisted. The steady-state numbers above are from a subsequent call against the now-warm daemon, which is the number that matters for a long-lived daemon serving a repeating agent loop.

## Tests

New `tests/unit/test_orient_agent_daemon.py` (23 tests) covering all items from the spec's TDD list: orient/agent `_from_map`==cold byte-identical (unit), warm==cold end-to-end via a real spawned daemon, TRAP A (direct unit coverage of the trap-detection logic + an end-to-end client-discard proof via a mocked unreliable daemon response), TRAP B (unit + end-to-end), stale-invalidation on an added file (orient and agent), fail-open when the daemon is down or returns an error, native-only + `--gpu-device-ids` routing gates, response-cache bounded/oversized-skip, concurrent requests against a real daemon, cross-command key isolation (orient vs agent), and forced-off-in-CI inheritance.

Fixed 8 pre-existing tests across 6 files whose `monkeypatch.setattr(repo_map, "build_context_render", ...)` mocks were silently bypassed by the refactor (the seam moved to `build_context_render_from_map`) — retargeted to the new seam, preserving each test's original intent (one test needed an additional `build_repo_map` mock to keep an intentionally-raising `_imports_and_symbols_for_path` patch scoped to the DAR-specific call it was testing, since the initial scan is no longer hidden behind a full-mock boundary). Updated two content-pinned tests (`docs/CONTRACTS.md`'s `_DAEMON_RESPONSE_CACHE_SCOPE` string and its `test_session_daemon_lifecycle` pin) to include `/orient/agent`.

## Gates

- `ruff check src/tensor_grep tests`: **PASS**
- `ruff format --check --preview src/tensor_grep tests`: **PASS**
- `mypy src/tensor_grep`: **PASS** (79 source files, no issues)
- `pytest` (full suite, `tests/unit` + `tests/integration` + `tests/e2e`, ~4800 tests, dev+ast+semantic extras to match CI's `test-python` job exactly): **PASS** except one pre-existing, confirmed-unrelated flaky test (`test_trust_parity.py::test_audit_help_does_not_fall_through_to_search` — a terminal-width-sensitive Click help-text wrap; zero diff overlap with this PR, passes in isolation, fails/passes inconsistently by run composition on `main` too)

**Opus-gate PENDING (session_daemon surface).**

## Test plan

- [x] TDD test suite (23 tests) green
- [x] Full existing suite green (minus the one confirmed-unrelated pre-existing flake)
- [x] `ruff check` / `ruff format --check --preview` / `mypy` all green
- [x] Dogfooded the real shipped `tg` binary: cold-vs-warm byte parity + latency numbers for both `orient` and `agent`
- [ ] Mandatory Opus adversarial-security gate (session_daemon surface)


oimiragieo:
## Opus-gate FIX-FIRST addressed: literal-seed rescue restored on the cold agent path

Pushed `75ab2c5`. The gate correctly caught a real recall regression — the refactor had routed **cold** `build_agent_capsule` through the rescue-less `_collect_capsule_call_site_evidence_from_map`, making the rescue-equipped `_collect_capsule_call_site_evidence` dead code and letting the internal `daemon_evidence_unreliable` sentinel leak into cold output.

### Before → after (cold call-site-evidence wiring)
- **Before (regression):** cold `build_agent_capsule` → `build_agent_capsule_from_map` → `_collect_capsule_call_site_evidence_from_map` → `build_symbol_blast_radius_from_map` (**no literal-seed rescue**). Out-of-window callers silently dropped; sentinel could leak to cold `--json`.
- **After (fix):** a private `_rescue_call_site_evidence` flag on `build_agent_capsule_from_map`:
  - **Cold** (`build_agent_capsule` passes `True`) → `_collect_capsule_call_site_evidence` → the FS-backed `build_symbol_blast_radius` **rescue wrapper** (`_literal_symbol_seed_files` → rebuild w/ `extra_files` → re-run). Recovers out-of-window callers exactly like pre-PR `main`; hardcodes the sentinel to `False`.
  - **Warm/daemon** (`session_store` dispatch, default `False`) → keeps the single-cached-map win via the rescue-less `_from_map` collector, stamps `daemon_evidence_unreliable` on an untrustworthy truncated no_match; client discards on the sentinel and **falls back to the now-genuinely-rescue-equipped cold path**.
  - Ranking + `suggested_scope` (TRAP B) still shared through `_from_map` — only evidence collection differs. Fixed the false "never on the cold path" comment.

### New recall tests (bidirectional oracle)
- `test_cold_call_site_evidence_collector_rescues_out_of_window_def` — rescue-equipped (cold) **recovers** the in-window caller on a def-out-of-window truncated map; rescue-less (warm) returns empty + `unreliable=True`.
- `test_cold_agent_capsule_recovers_out_of_window_callers_and_omits_sentinel` — end-to-end cold recovers the caller **and** omits the sentinel. Verified to FAIL under the pre-fix routing (0 callers + sentinel present) and PASS after (1 caller + no sentinel).
- `test_daemon_evidence_unreliable_never_leaks_into_direct_cold_agent_json` — direct-cold `tg agent --json` on a truncated repo never carries the sentinel.
- Existing warm==cold parity + TRAP A + TRAP B + fall-to-cold-on-sentinel tests all still pass (26 tests in the suite).

Note: native ranking cannot surface an out-of-window symbol as the primary target (its def is unscanned — `_exact_query_match_primary_symbol` falls back to the same capped `rm["symbols"]`), so the end-to-end recall test forces the primary via the ranking seam (mirroring how an LSP cross-project resolution would); the REAL rescue collector then runs against the REAL FS. The collector-level oracle is the direct, ranking-independent proof.

### Sentinel confirmed no longer leaks to cold
Real-binary dogfood: cold `tg agent . "build_agent_capsule" --max-repo-files 40 --json` → exit 2 (truncated), valid capsule, `daemon_evidence_unreliable` absent. Installed-code dogfood: cold rescue collector recovers `aaa_caller.py` from a def-out-of-window truncated scan.

### 4-gate
- `ruff check src/tensor_grep tests`: **PASS**
- `ruff format --check --preview src/tensor_grep tests`: **PASS**
- `mypy src/tensor_grep`: **PASS** (79 files)
- `pytest` (unit + integration + e2e, dev+ast+semantic extras, foreground): **PASS** except the pre-existing terminal-width `test_audit_help_does_not_fall_through_to_search` flake (passes at `COLUMNS=200`; zero diff overlap — my change touches neither the `audit` command nor `review-bundle`).

Opus-gate re-review requested.

PR #556, opened 2026-07-12
fix(security): canonicalize exec parent (8.3/junction/\?\) before apply_policy to close fail-open edges (#126)

oimiragieo:
## Summary

Closes the `apply_policy` Windows canonicalization fail-open edges tracked as a fast-follow
in commit `e10c91d` (H2, #509): *"the parent-canonicalization hardening (8.3/junction/\\?\
edges) is intentionally NOT applied here -- it is a tracked fast-follow."*

- **The fail-open edge, cited**: `src/tensor_grep/cli/apply_policy.py:647` builds
  `resolved_path = Path(os.path.abspath(resolved_executable))` -- deliberately **lexical only**
  (never `Path.resolve()`), by design: the `#453` defense requires a repo-local shadow that is
  itself a symlink/junction to be judged by where it lexically sits, never by where it points
  (see the docstring on `_abspath_beneath_or_equal`, `apply_policy.py:414-439`). That lexical-only
  comparison is bypassable in the *other* direction. All three of the task's named edges were
  confirmed fail-open on `main` (empirically verified on this Windows dev box, not just read from
  the code):
  1. **8.3 short name** (`PROGRA~1` vs `Program Files`) -- `os.path.abspath()` does not expand it;
     `_abspath_beneath_or_equal`'s `normcase(normpath(...))` string compare then sees two
     unrelated strings for the identical on-disk location.
  2. **NTFS junction alias** -- `os.path.abspath()` does not traverse a junction; a repo-local
     shadow reached via a junction alias that lives outside the lexical `repo\...` string escapes
     the check even though it is the same file.
  3. **`\\?\` / `\\?\UNC\` extended-length prefix** -- verified empirically that *neither*
     `os.path.abspath()` *nor* `Path.resolve()` strips this prefix; an untrusted policy's
     `lint_cmd`/`test_cmd` can spell `argv[0]` with the prefix explicitly (`shutil.which()`
     returns an already-path-shaped `argv[0]` unchanged, no PATH search, no canonicalization), so
     it survives all the way to the confinement comparison and never string-matches the
     unprefixed root.

## The fix

- `_canonicalize_exec_parent(executable_path)` (new, `apply_policy.py`) resolves **only**
  `executable_path.parent` via `Path.resolve(strict=True)` -- never the full path, which would
  dereference the leaf and reopen `#453`. Verified empirically on Windows: this single call both
  expands 8.3 short-name components at every level of the parent chain
  (`GetFinalPathNameByHandleW` returns the filesystem's own canonical long name directly -- a
  second `GetLongPathNameW` pass was not needed) and fully traverses NTFS junctions in the parent
  chain, in one step. Returns `None` on any resolution failure; callers must fail closed.
- `_abspath_beneath_or_equal()` now strips a `\\?\` / `\\?\UNC\` prefix (new
  `_strip_extended_length_prefix()` helper) from both operands before the `normcase`/`normpath`
  comparison.
- `_run_policy_command()` denies when **either** the raw lexical form **or** the canonicalized
  form lands beneath the untrusted root (an OR, not a replacement) -- the two conditions cover
  disjoint gaps (raw catches "spelled in-repo", canonical catches "resolves in-repo but spelled
  differently"), so this can only ever deny a *superset* of what either check denies alone; a
  legitimate outside-root binary fails both and is unaffected. Canonicalization failure
  (`_canonicalize_exec_parent` returns `None`) fails closed -- denied, never allowed through on
  the raw comparison alone.

## Per-edge tests (`tests/unit/test_apply_policy.py`)

All new tests proven **RED against pre-fix `main`** (via `git stash` on just the source file) and
**GREEN with the fix** -- see the bidirectional-oracle section added above
`test_strip_extended_length_prefix_removes_prefix`.

| Edge | Test(s) | How it's real |
|---|---|---|
| `\\?\` prefix | `test_run_policy_command_rejects_extended_length_prefix_shadow_inside_repo` | Fully real end-to-end: the policy's own command string embeds a `\\?\`-prefixed absolute path to a real repo-local shadow executable; no mocking of the resolution mechanism. |
| `\\?\` prefix (unit) | `test_strip_extended_length_prefix_*`, `test_abspath_beneath_or_equal_matches_extended_length_prefixed_spelling`, `..._no_false_positive` | Pure string / path-comparison unit tests. |
| NTFS junction | `test_run_policy_command_rejects_junction_alias_shadow_inside_repo` | Fully real: a real `mklink /J` junction (no admin needed) aliases the repo from outside it; `shutil.which` is mocked only to return the aliased path (simulating what a real PATH/cwd search could surface). |
| NTFS junction (unit) | `test_canonicalize_exec_parent_resolves_real_junction` | Direct unit test against a real junction. |
| 8.3 short name (unit, real) | `test_canonicalize_exec_parent_expands_real_8dot3_short_name` | Real mechanism proof against the OS's own always-present `C:\PROGRA~1` alias for `C:\Program Files` -- forcing **fresh** 8.3 generation on a throwaway test directory needs `fsutil 8dot3name set <path> 0`, which needs an elevated volume handle (verified `Error 5: Access is denied` without admin on this dev box), so a fully-real *end-to-end* repro isn't available unprivileged. |
| 8.3 short name (wiring) | `test_run_policy_command_rejects_8dot3_alias_shadow_inside_repo` | `_run_policy_command`'s OR-guard wiring, with `_canonicalize_exec_parent` monkeypatched to return exactly what the (separately, really proven) mechanism resolves an 8.3 alias to. |
| `#453` preserved | `test_canonicalize_exec_parent_does_not_dereference_the_leaf_symlink` + the existing `test_run_policy_command_rejects_symlink_shadow_resolving_outside_cwd` (unmodified, still green) | Proves the parent-only-resolve design never dereferences the leaf. |
| Fail-closed | `test_canonicalize_exec_parent_fails_closed_on_nonexistent_parent`, `test_run_policy_command_denies_when_exec_parent_canonicalization_fails` | Canonicalization failure denies, never falls back to allow. |
| No regression (legitimate cases) | `test_run_policy_command_resolves_relative_executable_to_absolute` (fixture updated), `test_run_policy_command_allows_junction_alias_that_resolves_outside_repo` | A real outside-root binary, and one reached via a *legitimate* outside-repo junction alias, both still run. |

One pre-existing test's fixture (`test_run_policy_command_resolves_relative_executable_to_absolute`)
needed a real (not merely mocked-path) trusted-bin directory, since the new canonicalization step
genuinely touches the filesystem (`Path.resolve(strict=True)`) where the old code was pure string
manipulation -- the mock previously pointed `shutil.which()` at a directory that never existed on
disk, which real `shutil.which()` never does.

## 4-gate status

All green on the touched files and the full tree:

- `ruff check src/tensor_grep tests` -- pass
- `ruff format --check --preview src/tensor_grep tests` -- pass (336 files already formatted)
- `mypy src/tensor_grep` -- pass
- `pytest -q` -- `tests/unit/test_apply_policy.py`: **66/66 passed**. `tests/unit/test_mcp_server.py`
  policy/apply/rewrite subset: **39/39 passed**. Full `tests/unit`: **4310 passed, 2 failed, 13
  skipped** in 630s -- the 2 failures are `test_retrieval_dense.py::TestRealFetchedModel` (real
  fetched-model dimensionality/ranking tests), confirmed pre-existing and unrelated via an empty
  `git diff --stat <branch-point> -- tests/unit/test_retrieval_dense.py
  src/tensor_grep/core/retrieval_dense.py`.

## Opus-gate PENDING

`apply_policy` is a mandatory-Opus-gate surface per `AGENTS.md` -- this PR is a draft pending that
adversarial review before merge.

### DIFF
--- a/src/tensor_grep/cli/apply_policy.py
+++ b/src/tensor_grep/cli/apply_policy.py
@@ -645,6 +645,29 @@ def _run_policy_command(name: str, command: str, cwd: Path, timeout: int) -> dic
             detail=f"{name} command executable {(argv[0] if argv else command)!r} was not found on PATH.",
         )
     resolved_path = Path(os.path.abspath(resolved_executable))
+
+    # #126 (Opus re-gate, 4th same-class edge -- UNC / network-share smuggling): a UNC spelling of
+    # the executable -- \\host\share\..., the loopback admin-share \\127.0.0.1\C$\...\<repo>\evil.cmd
+    # / \\localhost\C$\..., or the \\?\UNC\... extended form -- names the SAME on-disk in-repo shadow
+    # through the network/admin-share namespace, which Path.resolve() does NOT map back to its C:\
+    # drive-letter form (verified empirically). So _canonicalize_exec_parent returns a still-UNC
+    # parent and NEITHER the raw-stripped nor the canonical beneath-compare starts with the local
+    # repo-root string -- the OR-guard below would pass and the shadow would spawn (reproduced live
+    # end-to-end by the adversarial gate on a box where the C$ admin share is reachable). A
+    # UNC-spelled executable can never be confined to a LOCAL drive-letter repo root by a string
+    # comparison, and a legitimate local tool always resolves to a drive-letter path (C:\..., and
+    # \\?\C:\... -> C:\... after the ext-length strip -- NOT a UNC prefix), so refuse any UNC
+    # executable path outright. Fail closed, before the beneath-guard.
+    stripped_exec = _strip_extended_length_prefix(str(resolved_path))
+    if stripped_exec.startswith("\\\\"):
+        return _command_result(
+            passed=False,
+            detail=(
+                f"{name} command executable {argv[0]!r}: refusing a UNC/network-share "
+                f"executable path {resolved_path} (cannot be confined to the local repo root)."
+            ),
+        )
+
     try:
         untrusted_root = cwd.resolve()
     except OSError:
--- a/src/tensor_grep/cli/apply_policy.py
+++ b/src/tensor_grep/cli/apply_policy.py
@@ -386,8 +386,33 @@ def _path_is_under(path: Path, root: Path) -> bool:
         return False
 
 
+def _strip_extended_length_prefix(path_str: str) -> str:
+    r"""Strip a Windows ``\\?\`` / ``\\?\UNC\`` extended-length prefix for a STRING
+    comparison (#126).
+
+    The prefix opts a path OUT of Windows' normal path processing (component
+    normalization, 8.3 lookup, ``MAX_PATH`` checks) and is preserved verbatim by both
+    ``os.path.abspath`` and ``Path.resolve()`` -- verified empirically on this box:
+    neither adds nor removes it. An untrusted policy's ``lint_cmd``/``test_cmd`` can
+    spell its ``argv[0]`` with this prefix explicitly; since ``shutil.which()`` returns
+    an already-path-shaped ``argv[0]`` unchanged (a path containing a separator skips
+    the PATH search and is returned as-is after an access check), the prefix survives
+    all the way to the confinement comparison in ``_abspath_beneath_or_equal``. Left
+    unstripped, a prefixed spelling of a location and an unprefixed spelling of the
+    identical location never string-match, so the beneath-or-equal check silently
+    returns False (not beneath) for a path that IS beneath -- fail OPEN. Stripping it
+    here, ahead of ``normcase``/``normpath``, closes that gap for both operands
+    uniformly. A no-op on POSIX and on any path that never had the prefix.
+    """
+    if path_str.startswith("\\\\?\\UNC\\"):
+        return "\\\\" + path_str[len("\\\\?\\UNC\\") :]
+    if path_str.startswith("\\\\?\\"):
+        return path_str[len("\\\\?\\") :]
+    return path_str
+
+
 def _abspath_beneath_or_equal(path: Path, root: Path) -> bool:
-    """True if ``path`` is ``root`` itself or nested anywhere beneath it (any depth).
+    r"""True if ``path`` is ``root`` itself or nested anywhere beneath it (any depth).
 
     Deliberately independent of ``_path_is_under()``/``Path.relative_to()``: the
     shadow-executable guard in ``_run_policy_command`` must run this check on an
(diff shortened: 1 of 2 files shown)


oimiragieo:
## Opus re-gate response: 4th same-class edge (UNC / network-share) fixed

The gate's FIX-FIRST finding was correct. Added commit `f1f4135` (rebased onto `main@35868fa`, which carries the #558 doc-format hotfix).

**Root cause (confirmed empirically on this box)**: `Path.resolve()` does not map the `\host\C$\...` admin-share namespace back to `C:\...`, so `_canonicalize_exec_parent` returns a still-UNC parent and neither the raw-stripped nor the canonical `_abspath_beneath_or_equal` compare starts with the local drive-letter repo-root string -> the OR-guard passes -> the in-repo shadow spawns.

**Fix**: right after `resolved_path` is built, before the beneath-guard, refuse any executable whose path (after stripping a `\?\` / `\?\UNC\` extended-length prefix) is a UNC path (starts with `\`). A UNC-spelled executable can never be confined to a LOCAL drive-letter repo root by a string comparison; a legitimate local tool always resolves to a drive-letter path (`C:\...`, and `\?\C:\...`->`C:\...` after the strip), so local invocations are unaffected. Fail closed.

**RED->GREEN proof** (`git stash` of just the source fix, tests kept): the 3 admin-share spellings of a real in-repo shadow each reached the mocked `subprocess.run` pre-fix (SPAWNED -> `AttributeError: 'NoneType' ... returncode`), and all pass post-fix:

| Test | pre-fix | post-fix |
|---|---|---|
| `..._rejects_unc_admin_share_shadow_inside_repo[\127.0.0.1\C$\]` | FAIL (spawned) | PASS |
| `..._rejects_unc_admin_share_shadow_inside_repo[\localhost\C$\]` | FAIL (spawned) | PASS |
| `..._rejects_unc_admin_share_shadow_inside_repo[\?\UNC\127.0.0.1\C$\]` | FAIL (spawned) | PASS |
| `..._rejects_unc_share_even_when_outside_repo` | FAIL | PASS |
| `..._allows_local_drive_letter_tool_not_flagged_as_unc` (no-regression) | PASS | PASS |

`tests/unit/test_apply_policy.py`: **71 passed** (66 prior + 5 new).

**4-gate**:
- `ruff check .` (whole repo) -- PASS
- `ruff format --check --preview .` (whole repo) -- CI-clean. The only local flags are 4 **untouched** docs (`docs/architecture.md`, `docs/planning/PROJECT_PLAN.md`, `docs/plans/backlog-100/cluster-124-evidence-signing.md`, `docs/plans/design-tensor-grep-2026-04-29_01-45-34.md`) that are **LF-in-index / CRLF-in-my-Windows-worktree** -- an EOL-only false-alarm (proven: normalizing CRLF->LF makes all 4 pass), and CI checks them out LF on Linux. My two changed files pass.
- `mypy src/tensor_grep` -- PASS (80 files)
- `pytest tests/unit` -- **4352 passed, 2 failed, 14 skipped**. The 2 failures are `test_retrieval_dense.py::TestRealFetchedModel` (`No module named 'model2vec'` -- the `semantic` extra is not installed in this venv); pre-existing + environment-only, unrelated to this change (empty `git diff --stat` for both dense files).

Opus-gate PENDING (focused re-gate on the UNC edge).

PR #557, opened 2026-07-12
fix(count): --count-matches refuses cleanly instead of silently under-counting when rg is unresolvable (#121)

oimiragieo:
## Summary

Task #121 asked for `tg --count-matches` to degrade gracefully (like `--json`/default search)
instead of hard-erroring when ripgrep (`rg`) is fully unresolvable. **The literal "hard-errors"
premise does not reproduce on current `main`** -- I verified this live before building anything.
The actual bug is a different, more severe member of the same Backend Fail-Closed Contract family:
**`--count-matches` silently returns the WRONG number** (exit 0, no visible signal) instead of
crashing or refusing. This PR fixes that, at both front doors that can reach it.

## Current vs. fixed behavior (verified live, real binary, not CliRunner)

Fixture: a 3-line file where line 1 contains `foo` **three times** and line 3 contains it once
(4 total occurrences, 2 matching lines).

| Invocation | Before | After |
|---|---|---|
| `tg --count-matches foo file` (rg unresolvable) | prints `2`, exit 0 -- **silently wrong** (real answer is 4; `2` is `-c`'s line-count, not `--count-matches`'s occurrence-count) | `Error: --count-matches reports ripgrep's per-occurrence match count, which requires the 'rg' binary; ...`, **exit 2** |
| `tg search foo file --count-matches --json` (rg unresolvable) | `{"total_matches": 2, "routing_backend": "NativeCpuBackend", ...}` -- **silently wrong**, and via a **second, independent code path** (native-binary delegation) | `{"ok": false, "error": "count_matches_requires_ripgrep", "detail": "..."}`, exit 2 |
| `tg --count-matches foo file` (rg available) | `4` (correct) | unchanged: `4` (correct) |
| `tg -c foo file` (rg unresolvable) | `2` (correct -- this is `-c`'s actual contract) | unchanged: `2` (correct) |

## Root cause

`--count-matches` reports ripgrep's **per-occurrence** count (every match on a line counts
separately); `-c`/`--count` reports the **per-line** count (one per matching line, regardless of
how many times the pattern occurs on it). Every engine that can serve a query without `rg` is
line-granular only and cannot tell the two apart:

- `rust_core/src/backend_cpu.rs:522` (`count_file_memmem`) and `:559` (`count_file_regex`), the
  Rayon-parallel native count fast path used by both the PyO3 extension (`RustCoreBackend`) and
  the compiled standalone `tg` binary, says so explicitly in its own comment: *"For ripgrep parity
  on count matches, we count MATCHING LINES, not total occurrences."*
- `src/tensor_grep/backends/rust_backend.py:143-156` only sets `count_only` from `config.count`
  (`-c`), never `config.count_matches` -- so a `--count-matches`-only request falls through to the
  regular per-line `search()` path, which never emits more than one `MatchLine` per line
  (`rust_core/src/backend_cpu.rs:101-114`, `:164-186`; mirrored by the pure-Python fallback in
  `src/tensor_grep/backends/cpu_backend.py`).
- `src/tensor_grep/core/pipeline.py:263-266` (`count_rust_fast_path`) only fires for
  `config.count`, so a `--count-matches`-only request isn't caught there either; with `rg`
  unavailable it falls through to `rust_secondary_fast_path` (`pipeline.py:417-420`), the SAME
  backend the default/`--json` degrade already correctly uses.
- The CLI then derives a per-file count from the (line-granular) match list
  (`src/tensor_grep/cli/main.py:7206-7210`, plus the same fallback in
  `src/tensor_grep/cli/formatters/ripgrep_fmt.py:104-106`) -- silently reporting a line-count as if
  it were the requested occurrence-count.

This is confirmed **not a novel gap**: the native `--index` fast path already refuses
`count_matches` for the identical reason
(`rust_core/src/main.rs:5541`, `IndexFlagPolicy::Refuse`, and the explicit
`if *count_matches { violations.push("--count-matches"); }` check at `:5394`).

## Why refuse instead of "degrade" (mirroring --json/default, with one deliberate deviation)

`--json`/default search's degrade (`pipeline.py:417-420`, confirmed via
`tests/unit/test_pipeline.py::test_should_fallback_to_rust_when_ripgrep_missing`) is a
**like-for-like engine swap**: `RustCoreBackend`'s match set IS what tg's own aggregate model
already uses for a plain search (`total_matches` is line-based on the rg-JSON path too --
`ripgrep_backend.py:243-253`, "Counting stays one-per-matching-line ... so total_matches / parity
with the other backends is unchanged"). There is **no correctness gap** to paper over.

`--count-matches` has no such equivalent fallback -- I verified live that every rg-less engine is
architecturally line-granular (see Root cause above), so "degrading" would mean silently answering
a *different question* than the one asked. Per AGENTS.md's Backend Fail-Closed Contract ("never
silently wrong output") and the `IndexFlagPolicy::Refuse` precedent for this exact flag, I refuse
cleanly instead of shipping a plausible-looking wrong number. `-c`/`--count` is completely
unaffected -- its line-count contract is exactly what the fallback engines already provide
correctly, so it keeps using `count_rust_fast_path` / native delegation unchanged.

The refuse is implemented with the SAME "clean, structured, exit 2" mechanism `--json`/default
already use for other unserviceable requests (`_exit_search_error`, used elsewhere for
`unsupported_flag`/`invalid_regex`/`path_not_found`) -- not a bare exception. Note in passing: I
found that `Pipeline`'s existing `ConfigurationError` (e.g. the `--pcre2`-without-a-PCRE2-`rg`
case, `pipeline.py:193`) is genuinely **uncaught** by `cli/main.py`'s `search_command` and crashes
with a raw traceback (verified live, exit 1) -- that's a real, separate, pre-existing bug, out of
scope here, which is why I deliberately did NOT reuse `ConfigurationError`/`Pipeline` for this fix.

## The two front doors (miss one and it silently misroutes)

- `src/tensor_grep/cli/main.py` `search_command` (the primary Python Typer app) -- fixed at
  `main.py:6922-6952`, right after `RipgrepBackend()`/`is_available()`, before the expensive
  directory walk.
- `src/tensor_grep/cli/bootstrap.py` `_can_delegate_to_native_tg_search` (the pre-Typer argv-based
  fast path that can hand the whole request to the **compiled standalone `tg` binary**, bypassing
  Python's `Pipeline` entirely) -- **independently reachable**: `count_matches` was already
  excluded from `main.py`'s own `_NATIVE_TG_DELEGATION_DEFAULT_REQUIRED_FIELDS`
  (`main.py:1914`), but bootstrap's separate argv-based gate had drifted out of parity (the same
  class of gap its own `-e`/`-f` exclusion comment already calls out). Any `--count-matches`
  combined with a trigger flag (`--json`, `--ndjson`, `--cpu`, `--force-cpu`,
  `--gpu-device-ids`) delegated straight past the first fix. Fixed at `bootstrap.py:461-476` by
  adding `--count-matches` to `unsupported_flags`.

I found the second gap by dogfooding the real binary with `--json` added to the repro command --
the routing metadata (`"routing_backend": "NativeCpuBackend"`) gave it away.

## Tests (TDD: written first, confirmed RED without the fix via `git stash`, then GREEN)

`tests/unit/test_cli_modes.py`:
- `test_count_matches_refuses_cleanly_when_rg_unresolvable` -- plain mode, exit 2, no traceback,
  no silently-wrong `2`, message names `--count-matches`/`rg`/`--count`.
- `test_count_matches_refuses_cleanly_when_rg_unresolvable_json` -- `--json` mode, structured
  `{"ok": false, "error": "count_matches_requires_ripgrep", ...}` envelope.
- `test_count_matches_still_uses_ripgrep_when_available` -- bidirectional half: skips if `rg` isn't
  on the box, otherwise asserts the real occurrence count (`4`, not `2`) via a real subprocess
  invocation (matches `test_debug_passthrough_keeps_stdout_match_only`'s existing pattern, since
  plain `--count-matches` is eligible for raw rg passthrough, which bypasses `CliRunner`'s captured
  stream).
- `test_count_still_degrades_via_native_engine_when_rg_unresolvable` -- regression guard: `-c`
  keeps working correctly (`2`) with rg unresolvable.

`tests/unit/test_cli_bootstrap.py`:
- `test_count_matches_does_not_delegate_to_native_binary` -- unit-level on
  `_can_delegate_to_native_tg_search`, mirroring the existing `-e`/`-f` and `--rank`/`--bm25`
  exclusion tests; also asserts `-c`/`--count` keep delegating (unaffected).

## + `uv pip install model2vec` for the semantic-search integration test)

- `ruff check src/tensor_grep tests` -- **pass**
- `ruff format --check --preview src/tensor_grep tests` -- **pass** (auto-fixed 2 formatting nits
  in the new tests with `ruff format --preview` before this check)
- `mypy src/tensor_grep` -- **pass** (79 source files, no issues)
- `pytest -q` -- **pass**, 4527 passed, 30 skipped, 0 failed (726s; the one pre-existing
  environment gap, a `model2vec`-missing integration test, was resolved by installing the
  `semantic` extra rather than skipped, so this is a genuinely full green run, not a
  git-diff-confirmed-unrelated skip)

No Rust files were touched (root cause traced into `rust_core/src/backend_cpu.rs` and `main.rs`
for citation, but the fix itself refuses at the Python front doors rather than changing the native
engine's match granularity), so `cargo test`/`cargo build` doesn't apply here.

## Opus-gate PENDING (count/backend path)

This touches the search command's routing/refuse logic on both front doors (`main.py`'s
`search_command` and `bootstrap.py`'s native-delegation gate) -- flagging for the mandatory
adversarial security/backends gate per AGENTS.md before merge.

## Follow-up (not in this PR)

- The pre-existing `Pipeline` `ConfigurationError`-is-uncaught-by-`search_command` crash (e.g.
  `--pcre2` with no PCRE2-capable `rg`, `pipeline.py:193`) is a real, separate bug I found while
  verifying this fix's own error-propagation path was safe. Out of scope here; flagging for a
  follow-up task.
- `docs/CONTRACTS.md:41` still lists `--count-matches` as a validated compatibility row -- that
  remains accurate (the contract is preserved whenever `rg` is available); I deliberately did not
  add a new contract bullet for the refuse-when-`rg`-unresolvable edge case to keep this PR
  scoped to the bug fix, per this repo's convention of batching docs changes separately.

### DIFF
--- a/scripts/agent_readiness.py
+++ b/scripts/agent_readiness.py
@@ -269,12 +269,23 @@ def _search_flag_tokens_for_sweep(command: list[str]) -> set[str]:
     return tokens
 
 
+def _ripgrep_available() -> bool:
+    """Whether tg can resolve a ripgrep binary (TG_RG_PATH / PATH / bundled). Gates the
+    #121 tolerance in the flag sweep below: `--count-matches` needs rg for its per-occurrence
+    count, so both tg front doors refuse cleanly (structured exit 2) when rg is unresolvable
+    instead of emitting a silently-wrong line count."""
+    from tensor_grep.backends.ripgrep_backend import RipgrepBackend
+
+    return bool(RipgrepBackend().is_available())
+
+
 def validate_public_search_advertised_flag_sweep(
     _stdout: str, repo_root: Path, _expected_version: str
 ) -> None:
     if shutil.which("tg") is None:
         raise ReadinessError("could not resolve public tg command for search flag sweep")
 
+    rg_available = _ripgrep_available()
     probe_dir = repo_root / "artifacts" / "agent_readiness" / "public_search_flags"
     probe_dir.mkdir(parents=True, exist_ok=True)
     (probe_dir / "app.log").write_text("ERROR failed\nINFO ok\n", encoding="utf-8")
@@ -322,6 +333,18 @@ def validate_public_search_advertised_flag_sweep(
         )
         stderr = completed.stderr.strip()
         stdout = completed.stdout.strip()
+        # task #121: `--count-matches` needs rg for its per-occurrence count; both tg front
+        # doors refuse cleanly (structured exit 2) when rg is unresolvable rather than emit a
+        # silently-wrong line count. Accept that refuse here instead of flagging it as a sweep
+        # failure -- but ONLY when rg is genuinely absent (when rg IS available the probe must
+        # still exit 0/1 as before, so a real regression that breaks the flag still fails).
+        if (
+            label == "root-option-first-count-matches"
+            and not rg_available
+            and completed.returncode == 2
+            and "count-matches" in stderr.lower()
+        ):
+            continue
         if completed.returncode not in {0, 1}:
             failures.append(
                 f"{label}: exit={completed.returncode}, command={command!r}, "
--- a/src/tensor_grep/cli/bootstrap.py
+++ b/src/tensor_grep/cli/bootstrap.py
@@ -458,6 +458,23 @@ def _can_delegate_to_native_tg_search(search_args: list[str]) -> bool:
         "--regexp",
         "-f",
         "--file",
+        # task #121: `--count-matches` reports ripgrep's per-OCCURRENCE count, which the
+        # separately-compiled native binary's fallback engine cannot produce (it is
+        # LINE-granular only, same as the Python CPU/Rust fallbacks -- see
+        # rust_core/src/backend_cpu.rs's `count_matches`, "count MATCHING LINES, not total
+        # occurrences"). Without this exclusion, `--count-matches` combined with a trigger
+        # flag (`--json`/`--ndjson`/`--cpu`/`--force-cpu`/`--gpu-device-ids`) delegated
+        # straight to the native binary and silently returned a LINE count mislabeled as an
+        # occurrence count (verified live: a 3-occurrence line undercounted to 1, exit 0, no
+        # visible signal) -- bypassing cli/main.py's OWN identical exclusion entirely
+        # (`count_matches` is already in `_NATIVE_TG_DELEGATION_DEFAULT_REQUIRED_FIELDS`
+        # there; this outer argv fast path had simply drifted out of parity with it, the
+        # same two-front-doors gap the -e/-f exclusion above already guards against).
+        # Excluding it here routes to the full CLI instead, which refuses cleanly when rg
+        # is unresolvable and otherwise still uses rg (correct occurrence counts) when it
+        # is. `-c`/`--count` is UNCHANGED: its line-count contract is exactly what the
+        # native binary's fallback already provides correctly, so it keeps delegating.
+        "--count-matches",
     }
     unsupported_prefixes = ("--format=", "--lang=", "--replace=", "--regexp=", "--file=")
     if any(arg in unsupported_flags or arg.startswith(unsupported_prefixes) for arg in search_args):
--- a/src/tensor_grep/cli/bootstrap.py
+++ b/src/tensor_grep/cli/bootstrap.py
@@ -1056,6 +1056,16 @@ def main_entry() -> None:
         ) or _search_args_include_unbounded_broad_scan(passthrough_search_args)
         invalid_regex = _search_args_include_obviously_invalid_regex(passthrough_search_args)
 
+        # task #121 note: `--count-matches` is excluded from `_can_delegate_to_native_tg_search`
+        # (it needs rg for its per-occurrence count, which the native binary's fallback engine
+        # cannot produce), but the `_prefer_rust_first_search()` OR-branch below can STILL route
+        # a bare `--count-matches` search to the native binary when TG_RUST_FIRST_SEARCH=1
+        # (`--count-matches` is not a `_requires_full_cli` flag). That is SAFE and deliberately
+        # NOT special-cased here: the native binary self-refuses count_matches via
+        # `require_ripgrep_or_exit` (rust_core/src/main.rs) when rg is unresolvable -- a clean
+        # exit-2, never a silent wrong count. Locked by
+        # test_cli_bootstrap.py::test_rust_first_count_matches_refuses_via_native_self_guard so a
+        # future routing change to this branch cannot silently reopen the silent-wrong-count.
         if (
             not explicit_rg_json
             and native_binary is not None
--- a/src/tensor_grep/cli/main.py
+++ b/src/tensor_grep/cli/main.py
@@ -6920,10 +6920,40 @@ def search_command(
     from tensor_grep.io.directory_scanner import DirectoryScanner
 
     rg_backend = RipgrepBackend()
+    rg_is_available = rg_backend.is_available()
+    if config.count_matches and not rg_is_available:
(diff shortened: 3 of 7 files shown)


oimiragieo:
## FIX-FIRST remediation applied (Opus gate round 2)

Core logic unchanged (all 6 design checks passed). This commit updates the 3 governance surfaces that pinned the OLD `--count-matches` contract, per the repo rule that contract changes are updated in the SAME PR, plus rebases onto current `main`.

**1. Golden e2e (`tests/e2e/test_output_golden_contract.py`)** — the `count_matches_multi_file` / `count_matches_single_file` cases are now skipped when rg is unresolvable (mirrors the unit-test guard). `run_tg` hard-asserts exit 0, and both launchers (`python-m` TG_DISABLE_NATIVE_TG Python path + `native`) now refuse with exit 2 when rg is absent. Not re-snapshotted — the recorded `1`/`1` values are correct only because occurrence-count == line-count for the `hello` fixture *when rg is available*; with rg absent the output is a structured error, not a count.
- RED→GREEN proof (rg scrubbed from PATH): before = `assert 2 == 0` fail in `run_tg`; after = `4 skipped`. With rg available (normal) the cases still run and pass.

**2. Readiness (`scripts/agent_readiness.py`)** — the `root-option-first-count-matches` probe now tolerates the structured exit-2 `count-matches` refuse, but ONLY when `RipgrepBackend().is_available()` is False (a real regression still fails when rg IS available). Bidirectional unit tests added.
- RED→GREEN proof: tolerance branch disabled → `ReadinessError: ... root-option-first-count-matches: exit=2 ...` (the exact predicted failure); branch enabled → passes. The `rg-available` guard test proves the tolerance is narrow.

**3. Third native-delegation path locked** — the `_prefer_rust_first_search()` / `TG_RUST_FIRST_SEARCH` branch (bootstrap.py) can still route `--count-matches` to the native binary (it is not a `_requires_full_cli` flag). Verified SAFE: the native binary self-refuses via `require_ripgrep_or_exit` (rust_core/src/main.rs) — exit 2, never a silent count. Added a call-site comment + a dogfood lock-test (`test_rust_first_count_matches_refuses_via_native_self_guard`, uses a 3-occurrences-on-one-line fixture so a silent line-count fallback would print `1`; asserts a refuse, not `1` or `3`) so a future routing change can't silently reopen the bug.

**Rebase**: onto `origin/main` @ v1.67.0 (picks up #558 formatting fix + uv.lock; clean, no conflicts).

**4-gate (real venv, foreground)**: `ruff check .` pass · `ruff format --check --preview .` — my 7 changed files clean; the only whole-repo flags are 4 pre-existing docs (`docs/architecture.md`, `docs/planning/PROJECT_PLAN.md`, `docs/plans/backlog-100/cluster-124-evidence-signing.md`, `docs/plans/design-tensor-grep-2026-04-29_*.md`) that are `i/lf w/crlf` — a local Windows CRLF working-tree smudge (`.gitattributes` doesn't pin `*.md`), byte-identical to `main`, format-green on CI's LF checkout (as #558's identical-tree Formatting check passed). Zero `.md` files in this PR's diff · `mypy src/tensor_grep` pass (80 files) · `pytest tests/e2e/test_output_golden_contract.py tests/unit/test_cli_modes.py tests/unit/test_cli_bootstrap.py tests/unit/test_agent_readiness_script.py` = 633 passed, 2 skipped.

Opus re-gate PENDING (count/backend routing path).

PR #588, opened 2026-07-14
fix(gpu): fail-closed refuse on sidecar path for flags it cannot honor (#165)

oimiragieo:
## Summary

Fast-follow to #131 F3 (finding N1, flagged by the F3 Opus gate). F3 fixed the
CPU-fallback route's silent drop of `replace`/`only_matching`/`max_filesize`/
`color`/`no_ignore_vcs` when combined with `--gpu-device-ids`. A residual
remained on the **sidecar** route: `gpu_sidecar_search_payload` serializes all
5 flags into the sidecar's JSON request payload, but the Python sidecar's
`_gpu_search` (`sidecar.py`) reads none of them -- its `SearchConfig(...)`
construction has a fixed field allowlist that omits all 5. A
`--gpu-device-ids` search that routes to the sidecar (case-insensitive, a
regex pattern, `--invert-match`, etc. all do per `gpu_native_fallback_reason`)
with one of the 5 set silently ran without it -- exit 0, wrong output.

- Adds `gpu_sidecar_unhonorable_flag` + `exit_gpu_sidecar_flag_unhonorable`,
  mirroring F3's `gpu_cpu_fallback_unhonorable_flag` /
  `exit_gpu_cpu_fallback_flag_unhonorable` pattern (exit 2, name the flag).
- Unlike the CPU-fallback route (which can express `replace`/`only_matching`
  via `NativeSearchConfig`), the sidecar has **no** escape hatch for any of
  the 5 -- so all 5 are unhonorable here, not just the 3 the CPU-fallback
  gate checks.
- `--color never`/`auto` stay excluded as honorable no-ops (N3 parity, via
  the shared `gpu_color_is_unhonorable`).
- The common case (none of the 5 set) is untouched -- still routes to the
  sidecar unchanged, locked by the new test's first assertion.

## Design decision: guard placement (single chokepoint)

The guard lives **inside `handle_gpu_sidecar_search` itself**, not in either
`handle_gpu_search` variant. `handle_gpu_sidecar_search` and its sibling
helpers (`gpu_sidecar_search_payload`, `gpu_cpu_fallback_unhonorable_flag`,
etc.) are **not** `#[cfg(feature = "cuda")]`-gated -- they compile in both
configurations. I verified there are exactly 3 call sites into it:

1. cuda `handle_gpu_search`, via `gpu_native_fallback_reason` returning `Some`
2. non-cuda `handle_gpu_search`, via `explicit_gpu_sidecar_is_available()`
3. cuda `handle_gpu_native_search`'s empty-`gpu_device_ids` branch (not named
   in the task brief -- found while tracing call sites)

One guard at this single chokepoint covers all three, including the third
path the original brief didn't enumerate, rather than duplicating the check
at each call site (per "prefer ONE guard over two").

## Design decision I was unsure about: `line_number` is NOT gated (flagging for review)

`gpu_sidecar_search_payload` also serializes `line_number`, and the sidecar's
plain-text renderer always emits `file:line:text` -- it cannot suppress line
numbers either. I deliberately did **not** add `line_number` to the refuse
gate:

- `GpuSearchParams::line_number` collapses `cli.line_number &&
  !cli.no_line_number` into a single bool that defaults **false** whenever
  neither `-n` nor `-N` is passed -- i.e. the overwhelming common case. The
  existing shared test fixture (`gpu_params_for_patterns`) bakes in
  `line_number: false` as its own baseline value.
- Refusing whenever `line_number == false` would refuse nearly every
  sidecar-routed search (not just the rare ones carrying one of the 5 flags
  above), which directly breaks the "must still route the common case to the
  sidecar unchanged" requirement -- and the collapsed boolean can't
  distinguish "explicit `-N`" from "neither flag passed" to narrow it down.
- It's also a narrower, display-only mismatch (extra/missing line-number
  prefix) vs. the silently-wrong-**content** class the 5 flags above are (a
  dropped `--replace` transform, an unbounded `--max-filesize`, etc).

Documented in `gpu_sidecar_unhonorable_flag`'s doc comment as a known,
pre-existing gap. Flagging explicitly for the adversarial gate -- if this
read is wrong, the fix is a one-line addition to the same function, guarded
the same way.

## Verification

- RED confirmed cleanly: temporarily renamed `gpu_sidecar_unhonorable_flag`
  -> 10 `E0425 cannot find function` compile errors (every test assertion +
  the guard call site), then restored.
- `cargo test --bin tg`: **92/92 pass** (default, non-cuda).
- `cargo test --features cuda --bin tg`: **96/96 pass** (CUDA 13.1 toolkit
  available locally -- ran the real cuda-gated test set, not just the
  non-cuda-observable assertion).
- `cargo fmt -- --check`: clean.
- `cargo clippy -- -D warnings`: clean on both feature sets.

Scope: Rust-only, as directed (no `sidecar.py` changes -- reading the flags
there is the larger "option 1" the task explicitly deferred; this is the
smaller, more consistent fail-closed refuse).

## Test plan

- [x] `cargo test --bin tg` (92/92 pass)
- [x] `cargo test --features cuda --bin tg` (96/96 pass)
- [x] `cargo fmt -- --check` clean
- [x] `cargo clippy -- -D warnings` clean (default + cuda features)
- [ ] Mandatory adversarial Opus gate (fail-closed contract change) before merge

### DIFF
--- a/rust_core/src/main.rs
+++ b/rust_core/src/main.rs
@@ -2552,6 +2552,81 @@ mod tests {
         assert_eq!(gpu_cpu_fallback_unhonorable_flag(&params), None);
     }
 
+    // -- task #165 (fast-follow to #131 F3, Opus-gate finding N1): sidecar flag completeness -----
+    //
+    // F3 fixed the CPU-fallback route's silent drop of these 5 flags via
+    // `gpu_cpu_fallback_unhonorable_flag`. But `handle_gpu_sidecar_search` -- reached whenever
+    // `gpu_native_fallback_reason` (cuda) or `explicit_gpu_sidecar_is_available` (non-cuda) or the
+    // empty-`gpu_device_ids` branch of `handle_gpu_native_search` (cuda) routes here -- hands its
+    // `GpuSearchParams` to the Python sidecar via `gpu_sidecar_search_payload`'s JSON, and the
+    // sidecar's `_gpu_search` (`sidecar.py`) reads NONE of the 5 fields from that payload: its
+    // `SearchConfig(...)` construction has a fixed field allowlist that omits all of
+    // replace/only_matching/max_filesize/color/no_ignore_vcs. So `tg PAT --gpu-device-ids 0
+    // --ignore-case --replace X` (case-insensitive forces sidecar routing per
+    // `gpu_native_fallback_reason`) silently ran WITHOUT `--replace`, exit 0 -- same bug class as
+    // F3, one hop further down the same struct. This test would not even compile against a
+    // codebase missing `gpu_sidecar_unhonorable_flag` -- itself the RED signal, per F3's own
+    // precedent above.
+    #[test]
+    fn gpu_sidecar_unhonorable_flag_detects_each_of_the_five() {
+        let patterns = vec!["needle".to_string()];
+
+        // Preserve case: none of the 5 flags set must still route to the sidecar unchanged.
+        let params = gpu_params_for_patterns(&patterns);
+        assert_eq!(
+            gpu_sidecar_unhonorable_flag(&params),
+            None,
+            "the common case (no unhonorable flags) must still route to the sidecar unchanged"
+        );
+
+        // Unlike `gpu_cpu_fallback_unhonorable_flag`, the sidecar can express NEITHER `replace`
+        // NOR `only_matching` (it has no `NativeSearchConfig`-equivalent escape hatch) -- both
+        // must be unhonorable here even though they are honorable on the CPU-fallback route.
+        let mut params = gpu_params_for_patterns(&patterns);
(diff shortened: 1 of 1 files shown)


oimiragieo:
Closing unmerged. Dogfood-verify of v1.74.1 (#166) proved the rust handle_gpu_search / GpuSearchParams path this PR hardens is CLI-DEAD-CODE: bootstrap forces --gpu-device-ids to the Python CLI, and the Python Pipeline (pipeline.py:309-346) handles GPU via CuDFBackend/TorchBackend, never invoking the rust path. F3 (#131) shipped the sibling fix on the same dead path. Not shipping more dead-code hardening (YAGNI). The real user-facing GPU-flag handling lives in pipeline.py; tracked in #166.

PR #647, opened 2026-07-17
fix(daemon): default wall-clock deadline on the warm dispatch path (#200)

oimiragieo:
## Summary
- The warm session daemon's `_serve_session_request_from_payload` served `agent`/`orient`/`context_render`/`context_edit_plan` from an already-cached `repo_map` but never threaded a wall-clock budget into the post-map builders (`grep -n deadline session_store.py` was zero matches pre-fix) — the cold path's implicit 60s bound (`DEFAULT_AGENT_CLI_DEADLINE_SECONDS`) lives inside the cold branch only, so a **default** (no `--deadline` flag) warm-daemon request ran fully unbounded.
- Adds `WARM_DAEMON_DEFAULT_DEADLINE_SECONDS = 60.0` (matches the cold path's own default — a safe first cut, no new truncation for any response that already finishes inside budget) and threads `deadline_monotonic` into all 4 named dispatcher branches.
- **Deviation from the original verify (flagged explicitly, not silently absorbed):** `build_orient_capsule_from_map` did **not** already accept `deadline_monotonic`, unlike its 3 siblings (`build_agent_capsule_from_map` / `build_context_render_from_map` / `build_context_edit_plan_from_map`, which already did per #639/#642/#645). Extended it to accept + honor the parameter (bounds the snippet-building file-read loop + a final wall-clock catch-all mirroring `build_agent_capsule_from_map`'s existing `deadline_exceeded_at_return` pattern). The cold CLI wrapper (`build_orient_capsule`) is deliberately left untouched — `tg orient`'s cold path intentionally stays unbounded by default (see its own docstring / `--no-deadline` CLI help), so only the warm daemon path now supplies a value.
- Excludes `partial=True` responses from the daemon's response cache (`session_daemon._serve_daemon_response_with_cache`) so a deadline-truncated answer is never replayed to a later request that would have had a fresh budget to finish.

## Test plan
- [x] TDD: `tests/unit/test_session_daemon_default_deadline.py` (RED confirmed pre-fix, GREEN post-fix) — deterministic already-expired-deadline overrun stamps `partial=True` for all 4 commands (+ end-to-end exit-2 proof for `agent` via a real in-process daemon), partial responses excluded from the cache (recompute, not replay), normal in-budget responses unaffected (no spurious `partial`), and direct coverage of the new `orient_capsule.build_orient_capsule_from_map(deadline_monotonic=...)` parameter incl. a byte-identity golden-parity guard for `deadline_monotonic=None`.
- [x] Full regression sweep (~1400 tests) across every session/daemon/orient/deadline-touching test file: `test_orient_agent_daemon.py`, `test_orient_capsule.py`, `test_symbol_daemon_response_cache.py`, `test_symbol_daemon_autostart.py`, `test_session_*.py`, `test_repo_map_deadline.py`, `test_cli_deadline_coverage_gaps.py`, `test_cli_deadline_flag.py`, `test_mcp_server.py`, `test_agent_codemap_deadline_scale.py` (integration), docs-governance — all green, zero regressions.
- [x] `ruff check .`, `ruff format --check --preview .` (changed files clean; repo has 4 unrelated pre-existing Windows-CRLF-vs-preview-md drift files, untouched by this PR), `mypy src/tensor_grep` (strict) — all pass.
- [x] Confirmed no fast-path latency change: the deadline check is a single cheap `monotonic()` read per dispatcher branch plus existing checkpoint sites; a normal (<60s) response is unaffected.

Draft only, not to be merged without further review (Opus gate pending). #390 (the 5 symbol commands — defs/impact/refs/callers/blast_radius — plus blast_radius_render/blast_radius_plan/file_importers/context still running unbounded on the daemon path) remains open and out of scope for this PR.

### DIFF
--- a/src/tensor_grep/cli/orient_capsule.py
+++ b/src/tensor_grep/cli/orient_capsule.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 
 import fnmatch
 import json
+import time
 from pathlib import Path
 from typing import Any
 
@@ -775,6 +776,7 @@ def build_orient_capsule_from_map(
     render_profile: str = "compact",
     ignore: tuple[str, ...] = (),
     auto_deweight: bool = True,
+    deadline_monotonic: float | None = None,
 ) -> dict[str, Any]:
     """Task #108 (Tier-2 daemon moat): the map-based core of ``build_orient_capsule``, taking an
     already-built ``rm`` (e.g. the warm session daemon's cached ``repo_map``) instead of scanning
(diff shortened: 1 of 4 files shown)


oimiragieo:
## Opus adversarial gate (load-bearing: latency moat + cache correctness)

**Verdict: SHIP-WITH-NITS.** Full review probed 8 areas (cache-correctness race/logic, byte-identity/no-op-on-None claim, the "no new truncation under 60s" claim, orient's internal correctness/ordering, whether the deadline genuinely reaches a real check, missed dispatcher branches, thread safety, test quality) against the real diff (`git show 672a84c`). No FIX-FIRST blockers.

Findings, addressed in `6a31160`:
1. **Cache-guard coupling** (nit): `_serve_daemon_response_with_cache`'s new `if not response.get("partial")` guard is shared by the 5 symbol commands too. Verified it's a proven no-op for them today (they never emit `partial=True` on the warm path while #390 stays open) — documented so a future #390 fix doesn't have to rediscover this.
2. **Scoping** (non-blocking, called "a defensible judgment call"): the plain `context` command (distinct from `context_render`) was left out even though its builder (`build_context_pack_from_map`) already accepts `deadline_monotonic` — a near-free follow-up, deliberately NOT folded into this PR silently. Flagging for reviewer/CEO judgment on whether to fold in or file separately.
3. **Test coverage** (nit): added `test_orient_snippet_loop_breaks_mid_loop_not_just_pre_check`, mirroring `test_repo_map_deadline.py`'s proven fake-clock technique, to prove the orient deadline check cuts the loop short mid-iteration, not just pre-expired.

Everything else (cache-correctness, byte-identity, no-regression, real deadline enforcement, thread safety) came back clean with file:line citations.

PR #665, opened 2026-07-18
fix(security): uniform symlink-refusal + atomic-rename hardening across the JSON/key/audit-manifest writers (C4/#659 residual)

oimiragieo:
## Summary

The C4/#659 gate (already on `main`) hardened the evidence-bundle atomic write against a CWE-59
symlink attack with a **precheck + same-dir-temp + fsync + `os.replace`** pattern, and established
that `O_NOFOLLOW` is a **no-op on Windows** — the real cross-platform guard is the precheck +
same-dir-temp + rename shape, not the flag. This PR is the residual task: uniform-ize that exact
pattern across every sibling atomic writer in `cli/`, closing the gaps a fresh sweep found instead
of copy-pasting a fourth/fifth near-identical implementation.

**Approach**: one new shared primitive, `_index_lock.atomic_write_bytes` (+ a
`atomic_write_json` convenience wrapper for the common `json.dumps(payload, indent=2)` shape).
`_index_lock.py` already had zero internal-package imports and already hosted `replace_with_retry`
(used by all these siblings), so it's a safe, non-circular home. Every touched writer now
delegates to it; **no writer's serialization/output changed, only the write mechanics**
(atomicity + symlink-safety).

## Per-writer status

| Writer | Before this PR | After |
|---|---|---|
| `session_store._write_json_atomic` | Already had precheck+same-dir-temp+fsync+replace (the original C4 site). Gap: `O_NOFOLLOW` missing entirely; the `mode=None` branch used plain `open()` (no `O_EXCL` either). | Delegates to the shared helper — both branches now get `O_EXCL`\|`O_NOFOLLOW`(POSIX) uniformly. |
| `evidence_signing._write_private_key_atomic` | Already had precheck+same-dir-temp+`O_EXCL`+fsync+replace. Gap: `O_NOFOLLOW` missing. | Delegates to the shared helper (`mode=0o600`); its own precheck stays in front so the caller-visible exception is still `EvidenceSigningError`, not the helper's generic `OSError`. |
| `evidence_signing.generate_keypair`'s `.pub` sibling write | **Bare `write_text`** — no precheck, no atomicity, no fsync at all. | Own precheck (mirrors the private-key writer, so it raises `EvidenceSigningError` too) + delegates to the shared helper (`mode=0o644`). |
| `checkpoint_store._write_json_atomic` | **No `is_symlink()` precheck at all** — the biggest gap of the four named targets. Had same-dir-temp+fsync+replace but a pre-existing symlink at `metadata.json`/`index.json`/the discovery cache was silently *replaced* (not corrupted-through, but destroyed with no signal). | Delegates to the shared helper — gains the precheck + `O_EXCL`\|`O_NOFOLLOW`. |
| `audit_manifest._write_history_index` | **Bare `write_text`** — no precheck, no atomic temp+rename, no fsync at all — for the tamper-evident `.tensor-grep/audit/index.json` history chain. (`create_review_bundle`'s own `output_path` write was *already* hardened via `session_store._write_json_atomic` + its own precheck — left untouched, it transitively picks up the `O_NOFOLLOW` upgrade.) | Delegates to the shared helper. |
| `session_daemon._write_daemon_metadata_windows` | Not one of the 4 named targets, but found during the sweep: this Windows-specific copy of the pattern (needed so the ACL lockdown can run *between* temp-file creation and the token write — a separate, already-hardened concern, audit #81/#13) never gained the precheck. Carries the daemon's IPC secret token. | Added the precheck directly (kept **out** of the shared helper — routing it through would break the create-then-lock-then-write ordering the ACL fix depends on). `O_NOFOLLOW` doesn't apply here (Windows-only call site by construction). |
| `dogfood._write_json_atomic` | A **fourth** near-identical `_write_json_atomic` (found via the "any other sibling" sweep) — no precheck, no fsync, no `O_EXCL`, and a *predictable* (non-`uuid4`) temp filename. | Delegates to the shared helper via `atomic_write_bytes` (its serialization — `sort_keys=True` + trailing newline — differs from the common shape, so it stays outside `atomic_write_json` and passes its own precomputed bytes). |

**Left untouched, flagged for the gate**: `codemap.py::_atomic_write_text` (generates
`docs/code-map/*` pages) has the same missing-precheck shape, but it's a different risk class —
repo-documentation generation, not part of the session/evidence/checkpoint/audit trust chain the
C4 fix protects — so it was kept out of scope to keep this diff tight. `main.py`'s
`_write_json_refuse_symlink` (ruleset baseline/suppressions) was **not** touched — it already has
its own precheck + `O_NOFOLLOW` (in fact it's the codebase's existing reference for the
`getattr(os, "O_NOFOLLOW", 0)` idiom, cited in the new shared helper's docstring) and isn't part
of this writer family. The Rust-side `write_audit_manifest` TOCTOU is the separate,
already-tracked #110 — Python-side only per the task.

## Why not naive `O_NOFOLLOW`

`O_NOFOLLOW` is a documented no-op on Windows — verified empirically in this session (`hasattr(os,
"O_NOFOLLOW")` is `False` on this box). Relying on it alone would give false confidence on this
repo's primary dev platform. The uniform baseline is **`is_symlink()` refusal precheck (pre-resolve)
+ same-directory temp + fsync + `os.replace`**, which holds identically cross-platform, with
`O_CREAT|O_EXCL|O_NOFOLLOW` (`getattr`-guarded) on the temp file as POSIX-only
belt-and-suspenders on top.

## Testing

TDD throughout — a symlink-refusal test per newly-hardened writer, asserting the pre-existing
symlink is refused (`OSError`, or `EvidenceSigningError` at the two evidence-signing sites) and
its *target* is left untouched, plus a normal-write sanity test and (where relevant) an
overwrite-of-a-regular-file test. Two are genuine end-to-end tests through the *public* entry
point rather than the raw writer (`checkpoint_store.create_checkpoint`,
`audit_manifest.record_audit_manifest`), using a deterministic clock/uuid monkeypatch to avoid
any timing race. The shared primitive itself (`atomic_write_bytes`/`atomic_write_json`) has direct
unit coverage in `test_index_lock.py`: symlink refusal, normal write, overwrite-of-regular-file,
parent-dir creation, `O_CREAT`/`O_EXCL`/`O_NOFOLLOW` flag assertion (spied), mode
apply+reassert, default-mode-matches-`open()`, and a crash-before-fsync atomicity proof
(monkeypatched `os.fsync`, mirroring the existing C4 regression test's technique).

One pre-existing test (`test_session_cli.py::test_session_open_fsyncs_payload_before_atomic_replace`)
monkeypatched `session_store.replace_with_retry` directly; since the real write now happens inside
`_index_lock.atomic_write_bytes`, that patch point no longer intercepts anything (a per-module
`from X import name` binding, unlike the shared `os` module singleton the same test's `fsync` patch
relies on) — retargeted to patch `_index_lock.replace_with_retry` instead, same assertions,
verified passing. `test_evidence_signing.py`'s existing `os.open`-restrictive-mode spy test now
also observes the *new* `.pub` file's `os.open` call — rescoped to assert the private-key temp
(`0o600`, no group/other bits) and the `.pub` temp (`0o644`) separately, since blanket-asserting
"every `os.open` call is restrictive" is no longer the intended contract now the `.pub` write is
correctly *less* restrictive (it's public material).

**Results**:
- Ruff format (`--preview`) + ruff check: clean on all 14 touched files.
- `python -c "import ..."` sanity import: all 7 touched modules + `main`/`mcp_server` import cleanly.
- Targeted test files (14 touched + siblings): 261 passed, 4 skipped (POSIX-only, correctly
  skipped on this Windows box), 0 failed.
- Full regression sweep — every test file in `tests/unit/` that references any of the 7 touched
  modules (78 files, ~3000 tests): **2969 passed, 7 skipped, 5 deselected, 0 failed.** The 5
  deselected are pre-existing failures **unrelated to this change** — verified individually by
  `git stash`-ing this diff and re-running each on the clean `origin/main` (eae8ea9) baseline,
  where they fail identically: 4 in `test_mcp_server.py` (a `rewrite_plan`/`rewrite_apply`
  "embedded fallback when native binary is missing" cluster — this worktree has no compiled
  native `tg` binary, per the CPU-safe build rule) and 1 in `test_cli_modes.py` (a GPU
  device-inventory test, environment-dependent).
- Real-binary sanity: `tg --version` (via bootstrap) succeeds; `checkpoint create --json` and
  `evidence keygen --json` both exercise the changed write paths end-to-end through the real
  Typer CLI app with exit code 0 (including confirming the `.pub` file is written correctly).

## Confirmation

No output/return-value/behavior change for any writer's success path — only the write mechanics
(atomicity + symlink-safety) changed. On the fail-closed refusal path, a pre-existing symlink at
any of these destinations is now refused instead of silently replaced/destroyed (a deliberate,
intended fail-closed tightening — the whole point of this task).

Draft pending the independent Opus gate.

### DIFF
--- a/src/tensor_grep/cli/_index_lock.py
+++ b/src/tensor_grep/cli/_index_lock.py
@@ -1,11 +1,13 @@
 from __future__ import annotations
 
+import json
 import os
 import threading
 import time
 from collections.abc import Iterator
 from contextlib import contextmanager
 from pathlib import Path
+from typing import Any
 from uuid import uuid4
 
 _POLL_S = 0.02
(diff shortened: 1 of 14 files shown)


oimiragieo:
**Independent Opus gate: SHIP-WITH-NITS (safe to merge).** Correction to the PR body's "byte-for-byte unchanged" claim: serialization STRUCTURE is unchanged, but on Windows the writers that previously used text-mode writes now normalize **CRLF→LF** (the new binary helper). This is **functionally inert** — nothing hashes these files' raw bytes (the audit chain digests `_canonical_manifest_bytes`, re-serialized from the parsed object and line-ending-agnostic; index.json is never hashed; all readers use `json.loads`/`read_text` which normalize newlines), and it's arguably an improvement (consistent LF). Note: `daemon.json` keeps text mode, so `.tensor-grep/` mixes LF/CRLF cosmetically. Gate verified: primitive correctness (no TOCTOU/mode window), all 7 writers' output structure preserved, real-symlink refusal tests per writer, error contracts intact, no import cycle, scope calls all correct (esp. session_daemon's direct precheck preserving the ACL-lock-before-token-write ordering, audit #81/#13).

PR #666, opened 2026-07-18
fix: context-render/blast-radius-plan/blast-radius-render --max-files now bounds suggested_edits (broader B9/#661 flag-lie, #212)

oimiragieo:
## Summary

VERIFY-FIRST task for #212, a follow-up to B9/#661. Hypothesis: `tg context-render` and
`tg blast-radius-plan`/`blast-radius-render` advertise a `--max-files` cap flag but silently
ignore it for `edit_plan_seed.suggested_edits`, the same class of bug B9/#661 fixed for
`tg edit-plan`.

**Verdict: REAL, confirmed for all three commands** before writing any fix.

### Verification (before touching code)

1. **Source-level**: B9's own docstrings already documented the exact gap.
   `_capped_suggested_edits`'s docstring (repo_map.py) said context-render's "compact"/"llm"
   profiles get a downstream cap via `_compact_edit_plan_seed`, but the default **"full"
   profile has no cap at all**, and blast-radius-plan/render (which never call
   `_compact_edit_plan_seed`) have **no cap at all, in any profile**. All three were marked
   "deliberately left unbounded -- out of scope" for B9's PR.
2. **CLI contract**: all three commands DO advertise `--max-files` (typer help text:
   "Maximum files to include in the render bundle" / "...in the plan"), default 3.
3. **Live dogfood** (Python builders called directly against tensor-grep's own
   `src/tensor_grep`, 80 files, symbol `SearchConfig`):

   | command | `--max-files` | `files` | `suggested_edits` | distinct files touched |
   |---|---|---|---|---|
   | blast-radius-render | 1 | `[1]` | 73 | 40 |
   | blast-radius-render | 50 | `[42]` | 73 (**identical**) | 40 |
   | blast-radius-plan | 1 | `[1]` | 25 | 8 |
   | context-render (full) | 1 | `[1]` | 5 | 3 |

   `blast-radius-render` is the starkest case: `--max-files` has **zero** measurable effect
   on `suggested_edits` -- 1 vs 50 produce byte-identical counts.

## Fix

Mirrors B9's own opt-in pattern exactly -- no new CLI flags, no change to the low-level
`suggested_edits_max: int | None = None` default (any future/unnamed caller still gets the
unbounded pre-fix shape). Three builders now thread their own already-existing `--max-files`
value into `_attach_edit_plan_metadata`'s existing `suggested_edits_max` parameter:

- `build_context_render_from_map` -> `suggested_edits_max=max_files` (closes the "full"
  profile gap; "compact"/"llm" already truncated to the same bound one step downstream via
  `_compact_edit_plan_seed`, so this is a **provable no-op** for those two profiles -- see
  `test_context_render_compact_and_llm_profiles_stay_prefix_stable`).
- `build_symbol_blast_radius_plan_from_map` -> `suggested_edits_max=normalized_max_files`.
- `build_symbol_blast_radius_render_from_map` -> `suggested_edits_max=max_files`.

`related_spans` (the raw input to `suggested_edits`) is intentionally left unbounded,
matching edit-plan's own already-shipped B9 fix, which also never capped `related_spans` --
staying symbol-scoped to exactly what B9 capped.

## Before / after (dogfood re-run post-fix)

| command | `--max-files` | `suggested_edits` before | `suggested_edits` after |
|---|---|---|---|
| context-render (full) | 1 / 50 | 5 / 73 | 1 / 50 |
| blast-radius-plan | 1 / 50 | 25 / 73 | 1 / 50 |
| blast-radius-render | 1 / 50 | 73 / 73 | 1 / 50 |

## Tests

- Updated the two B9 regression tests in
  `test_edit_plan_max_files_bounds_suggested_edits.py` that pinned the OLD unbounded shape
  as "must not change" -- they now pin the corrected bounded shape (renamed
  `..._now_bounds_context_render_suggested_edits_in_every_profile` /
  `..._now_bounds_blast_radius_plan_and_render`).
- New file `test_context_render_and_blast_radius_max_files_bounds_suggested_edits.py`:
  RED/GREEN proof per command (monkeypatch `_capped_suggested_edits` to a raw passthrough to
  confirm the fixture produces more than the cap, then confirm the real fix truncates to a
  stable prefix), a larger-cap-never-inflates guard, a zero-related-spans guard, and the
  compact/llm double-truncation no-op check -- all parametrized across the three commands
  where practical.
- `18 passed` on the two files combined.
- Full regression sweep: every test file in the repo that references
  `build_context_render*`/`build_symbol_blast_radius_plan*`/`build_symbol_blast_radius_render*`
  / `context-render` / `blast-radius-plan` / `blast-radius-render` was run foreground and
  passes (agent_capsule, cli_deadline, cli_modes, mcp_server, mcp_caps, session/daemon,
  repo_map spans/targets/deadline/lexical-ranking, edit_plan_seed, blast_radius perf/mermaid,
  trust_planning, validation_commands, token_budget, semantic_provider_navigation,
  bakeoff/benchmark/profiling harnesses, agent_readiness/success, js/ts/rust advanced
  resolution, caller/import span targeting). A handful of pre-existing failures were found
  (AST rewrite/apply "embedded fallback" tests in `test_mcp_server.py`, and a GPU-inventory
  test in `test_cli_modes.py`) -- confirmed via `git stash` that all of them fail identically
  on unmodified origin/main code, unrelated to this diff (different subsystem: AST
  rewrite/apply native-binary resolution and local GPU device inventory, neither touched by
  this fix).
- `ruff format --preview` and `ruff check` clean on all three changed files.
- Import sanity confirmed (`tensor_grep.__file__` resolves into the worktree, `cli.main`/
  `cli.repo_map`/`cli.mcp_server` import cleanly).

## Scope discipline

- No CLI flag added or removed; `--max-files` was already registered and documented on all
  three commands.
- No change to the low-level opt-in default (`suggested_edits_max=None` everywhere except
  the four now-named callers).
- `related_spans` deliberately left unbounded, matching edit-plan's own precedent.
- Diff is symbol-scoped: 3 keyword-argument additions at existing call sites, plus docstring/
  comment updates so the "deliberately out of scope" notes B9 left behind don't mislead a
  future reader, plus the test updates described above.

Draft pending the independent Opus gate per this repo's Backend Fail-Closed / change-control
discipline (load-bearing repo_map.py surface).

## Test plan

- [x] `ruff format --preview` clean
- [x] `ruff check` clean
- [x] New + updated unit tests pass (18/18)
- [x] Broad regression sweep across every test file referencing the touched builders/commands
- [x] Live dogfood before AND after the fix, on tensor-grep's own repo
- [ ] Independent Opus gate review

### DIFF
--- a/src/tensor_grep/cli/repo_map.py
+++ b/src/tensor_grep/cli/repo_map.py
@@ -11340,14 +11340,21 @@ def _capped_suggested_edits(
     --max-files N`` grew ``suggested_edits`` unbounded despite the flag's documented meaning.
 
     ``max_edits`` is threaded in from ``_build_edit_plan_seed``'s ``suggested_edits_max`` parameter,
-    which defaults to ``None`` and is OPT-IN per caller (see that parameter's docstring): only
-    ``tg edit-plan``'s own builder (``build_context_edit_plan_from_map``) currently passes a real
-    value. ``tg context-render`` (own separate, pre-existing downstream cap via
-    ``_compact_edit_plan_seed``, but only for the compact/llm render profiles -- the default "full"
-    profile has no cap at all) and ``tg blast-radius-plan``/``tg blast-radius-render`` (no cap at
-    all) carry the identical flag-lie this function fixes, but stay on the ``None`` default and are
-    deliberately left unbounded -- out of scope for this PR; see
-    ``tests/unit/test_edit_plan_max_files_bounds_suggested_edits.py`` for the pins proving so."""
+    which defaults to ``None`` and is OPT-IN per caller (see that parameter's docstring).
+    ``tg edit-plan``'s own builder (``build_context_edit_plan_from_map``, audit B9/A18) was the
+    first to pass a real value. Audit #212 (a follow-up to B9/#661) found the identical flag-lie in
+    three sibling callers and closed it the same way: ``tg context-render`` (own separate,
+    pre-existing downstream cap via ``_compact_edit_plan_seed``, but only for the compact/llm render
+    profiles -- the default "full" profile had no cap at all) and ``tg blast-radius-plan``/``tg
+    blast-radius-render`` (no cap at all, in any profile) now ALSO opt in, passing their own
+    ``--max-files`` value through ``build_context_render_from_map``/``build_symbol_blast_radius_
+    plan_from_map``/``build_symbol_blast_radius_render_from_map`` respectively. See
+    ``tests/unit/test_edit_plan_max_files_bounds_suggested_edits.py`` and
+    ``tests/unit/test_context_render_and_blast_radius_max_files_bounds_suggested_edits.py`` for the
+    pins proving so. Any FUTURE caller of ``_build_edit_plan_seed``/``_attach_edit_plan_metadata``
+    that does not explicitly pass ``suggested_edits_max`` still gets the unbounded, pre-fix
+    behavior -- the low-level default stays ``None`` on purpose so a not-yet-written caller is never
+    silently capped."""
     if max_edits is None:
         return entries
     return entries[:max_edits]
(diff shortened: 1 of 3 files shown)


oimiragieo:
**Independent Opus gate: SHIP-WITH-NITS (safe to merge).** Disclosure the gate flagged: `--max-files` defaults to **3** (not None) on these commands, so with the flag UNSET, `suggested_edits` is now capped to 3 (was unbounded) — a visible default-output change, but **correct + B9-consistent**: (a) `build_context_edit_plan_from_map` already caps suggested_edits to 3 at default (repo_map.py:12562/12640), accepted since B9/#661; (b) `files` is already capped to 3 at default, so pre-fix suggested_edits (73 in dogfood) referenced files NOT in the bundle — incoherent, not a feature; (c) CONTRACTS.md:111/118 treats `--max-files` as a display budget that stays exit 0. Fix lives in the `_from_map` builders → also covers the warm-daemon/agent-capsule routes (correct — CLI-only capping would create the #107 warm-vs-cold divergence; `tg agent` unaffected, it reads edit_plan_seed for nav/validation only, never suggested_edits). `related_spans` left uncapped to match B9 exactly. Non-blocking follow-ups: a warm-daemon parity assertion + optional related_spans symmetric pass (tracked #216-batch).

PR #671, opened 2026-07-19
fix(agent): bound the vendored-subtree dedup loop's super-linear cost under deadline (real-workspace-scale residual of #669)

oimiragieo:
## Summary

An OLD-vs-NEW real-binary re-verify of #669 on a real ~50k-file/40-sibling-project workspace
falsified its magnitude claim: the fix works at small scale (1.24x overshoot at `--deadline 10`)
but a super-linear stage still engages at larger scale (>200s under cProfile at `--deadline 30`;
146s unprofiled at the 60s default = 2.43x). This PR localizes and bounds that stage.

**Root cause**: `_detect_vendored_subtrees`'s outermost-nested-chain dedup loop
(`orient_capsule.py`) called `_repo_map._path_is_relative_to(root / rel_dir, root / existing)`
inside an O(candidate_roots²)-shaped comparison. That helper does **two real `Path.resolve()`
filesystem syscalls** (Windows `nt._getfinalpathname`) **per pair**. The existing #220/#669
per-iteration deadline check correctly bounds the *iteration count*, but not each iteration's own
cost, which grows with `len(subtree_rel_roots)` — so on a workspace with thousands of
manifest-bearing directories, a handful of late, expensive iterations can blow tens of seconds
past `--deadline` between one checkpoint and the next.

## Reproducing-shape recipe (CPU-safe synthetic, not the real workspace)

`#669`'s own fixture (`manifest_heavy_repo`, 60 projects × 2 nested packages) does **not**
reproduce this: each project's own top-level manifest absorbs its nested `packages/pkgN`
children early in the depth-sorted dedup pass, so `subtree_rel_roots` never grows large. The
real workspace's shape has many *independent* manifest-bearing directories with **no common
absorbing ancestor** (a monorepo `packages/*/package.json` shape, or a vendored dependency
tree) — reproducing this required a synthetic where candidate roots stay independent:

- A wrapper directory (`deps/`) that is itself neither a STRONG-0 vendor name nor
  manifest-bearing, so its children never nest under a common accepted root.
- Note: a **literal** `node_modules/` does **not** reproduce this — it's in
  `repo_map._SKIP_DIR_NAMES` and is pruned before descent (zero cost, zero files reach
  `rm["files"]`). The synthetic uses a non-skip-pruned, non-absorbing wrapper instead.
- N independent `deps/pkg_NNNNN/{package.json,index.js,lib/index.js}` packages.
- A `.claude/skills/` tree (~200 skill dirs) and a handful of large files, per the shape
  guidance, to mirror the real workspace's composition (these did not end up dominating the
  cost — see Localization below).

## Before/after: unbounded single-call cost (the super-linear curve)

Measured directly (no CLI/subprocess overhead) via one call to `_detect_vendored_subtrees`,
fresh tree per data point (no cross-run cache warming):

| Candidates | Pre-fix (single call) | Post-fix (single call) | Speedup |
|---|---|---|---|
| 120 | 7.74s | — | — |
| 200 | 20.69s | 0.45s* | — |
| 304 | 40.86s | 0.45s | 90.8x |
| ~500 | 102.59s | 0.71s | 144x |
| 2400 | *(not safely measurable pre-fix — extrapolated ~40+ min)* | 5.87s | — |

*(pre-fix→post-fix scaling is ~quadratic: 200→304 candidates is 1.52x more candidates for 1.97x
more time, consistent with N².)*

## Localization (cProfile, cited)

Full-pipeline `cProfile` of `build_agent_capsule` at `--deadline 30` on a 200-candidate tree
(18M function calls, 30.06s total):

```
ncalls  tottime  cumtime  function
     2    0.089   27.811  orient_capsule.py:286(_detect_vendored_subtrees)   <- 92.5% of ALL wall-clock
 29878    0.081   26.733  orient_capsule.py:509(<genexpr>)                    <- the dedup any()
 29571    0.088   26.405  repo_map.py:1134(_path_is_relative_to)
 59847    0.108   20.721  pathlib resolve()
119702   18.177   18.177  {built-in method nt._getfinalpathname}              <- 61% of ALL wall-clock
     1    0.001    2.035  repo_map.py:6486(build_repo_map)                    <- collection stage: 7%
```

`_detect_vendored_subtrees` was called **twice** (call-1 in `agent_capsule.py`, call-2 inside
`repo_map._build_context_pack_from_map`'s own `auto_deweight` pass — the "cost paid twice" #669
already flagged) and together consumed 92.5% of total wall-clock. `build_repo_map` (the
collection/scan stage) was only 2.0s — not the bottleneck.

**Why `vendored_subtree_detection` doesn't show up in `assembly_stages_skipped` at 60s-scale**
(the task's key open question): confirmed both by code reading and empirically reproduced. Call-2
(`repo_map.py:8064`, inside `_build_context_pack_from_map`) correctly sets `payload["partial"] =
True` when its internal `deadline_hit` trips, but `agent_capsule.py`'s `skipped_assembly_stages`
list was only ever populated from call-1's *own, separate* flag — call-2's trip was invisible to
that list, surfacing only as the generic `partial: true` with **no name** at all. Reproduced: a
deadline=30 run on a 200-candidate tree returned `partial: true,
deadline_limit: {"deadline_exceeded": true}` with `assembly_stages_skipped` **absent entirely**.

## Fix sites

1. **`src/tensor_grep/cli/orient_capsule.py`** (primary fix) — replaced the resolve()-based
   nesting check in the dedup loop with a pure lexical `.parts`-prefix comparison. `rel_dir`/
   `existing` are both already relative to the same resolved `root` (built via a plain
   `.relative_to(root)`, no resolve() of its own), so no filesystem I/O is needed at all. Same
   fix shape this codebase already uses twice: this function's own STRONG-3
   code-file-membership test, and **PR #670**'s `_tier` helper (`repo_map.py:16057-16090`) —
   "resolve once / reuse / avoid re-resolving via `_path_is_relative_to`" is this codebase's
   established pattern for exactly this class of hot-loop cost.
2. **`src/tensor_grep/cli/repo_map.py`** — `build_context_render_from_map` gained a new,
   additive, default-`None` `deadline_hit` passthrough to `build_context_pack_from_map` (which
   already supported it). Every pre-existing caller (repo_map's own `build_context_render`,
   session_store's warm-path renders) is unaffected.
3. **`src/tensor_grep/cli/agent_capsule.py`** — a new, dedicated `context_pack_deadline_hit`
   flag is threaded into the render call and, when tripped, appended to
   `skipped_assembly_stages` as `"context_pack_assembly"` — deliberately a **different** name
   than `"vendored_subtree_detection"`, since the shared flag also covers the symbol-scoring and
   pagerank sibling loops inside that call; naming it generically avoids overclaiming precision
   this signal alone cannot support.

## Correctness (byte-identity)

- Ran the identical correctness check against `_path_is_relative_to`-based origin/main code and
  the new `.parts`-prefix code on a real nesting scenario (`third_party/` STRONG-0 root with a
  nested STRONG-1-manifest `pkgA/`): **identical output** — the nested manifest is correctly
  subsumed into one entry, both before and after.
- No-deadline-pressure byte-identity confirmed at both the unit level (far-future deadline vs.
  no deadline) and the integration level (ample `--deadline 60` vs. no `--deadline` flag at all
  — `suggested_ignore`/`target`/`alternatives`/`suggested_scope` all compared equal).
- Full related test surface: **494 passed, 1 skipped** (416 pre-existing agent/deadline/orient/
  context-pack unit tests + 67 session-daemon tests + 2 pre-existing #220 integration tests + 9
  new tests). `ruff format --preview` and `ruff check` clean on all touched files.

## New tests

- `tests/unit/test_agent_vendored_subtree_dedup_scale_222.py` (6 tests): a monkeypatch guard
  that fails immediately if the dedup loop ever calls the resolve()-based helper again;
  byte-identity on a real nesting scenario and with/without deadline pressure; the
  `deadline_hit` passthrough wiring, unit-level and end-to-end through `agent_capsule`.
- `tests/integration/test_agent_dedup_scale_sla_222.py` (3 tests): real-subprocess,
  real-wall-clock SLA (`wall <= deadline * 2` at two deadlines), outcome-agnostic on returncode
  per the #669 CI lesson (run 29671609290), double-timeout anti-hang protocol. Fixture sized to
  ~500 independent candidates — **verified this scale is genuinely discriminating**: a smaller
  200-candidate fixture was checked first and found to **pass even against the pre-fix code**
  (the old per-iteration check still bounds things tightly when the loop is cut off early in its
  cheap phase — the bug only bites once checkpoint granularity itself gets coarse deep in the
  curve), so the fixture was sized up until the pre-fix code was directly measured to fail
  (102.6s/call vs. an 8s×2=16s bound) before finalizing it.

## Honesty / scope notes for the gate

- **What was NOT fully reproduced end-to-end**: the exact real-workspace overshoot RATIOS
  (1.24x→2.43x+) were not replayed literally at 50k-file/thousands-of-candidates scale within
  the CPU-safe ≤120s-per-run bound — doing so would itself require a single profiled call
  costing many minutes (a direct, almost paradoxical consequence of the bug's own severity). The
  MECHANISM and its super-linear growth LAW are conclusively localized and quantified instead
  (clean near-quadratic fit across 4 measured points, cProfile citing exact `file:line`), and the
  fix is proven to eliminate the mechanism (144x speedup, zero resolve() calls remaining, per the
  monkeypatch guard).
- **Out of scope, left as residual/follow-up**: `_reverse_importers` / `_reverse_import_distances`
  (`repo_map.py`) have no deadline checks at all and are alias-collision-prone, but were measured
  to be a small fraction of wall-clock in every profile here (~3.4s of 30s, vs. 27.8s in the
  dedup loop) — not the dominant stage this task asked to localize and bound. The redundant
  double-call to `_detect_vendored_subtrees` itself (call-1 + call-2, both now cheap) was left
  as-is per scope discipline (the task asked to bound the identified stage, not eliminate
  redundant calls) — a natural follow-up, not required for this fix's target.
- The `--deadline` flag's default-60s cold-path behavior and the exit-code/partial contract are
  unchanged by this PR — this is a pure performance fix to an existing bounded-but-slow stage,
  plus an honesty/enumeration improvement, with no behavior change absent deadline pressure.

Pending independent Opus gate.


oimiragieo:
## Independent Opus gate: SHIP

A fresh Opus agent (no exposure to the author's reasoning, isolated worktree) adversarially reviewed this PR against the real code and rendered **SHIP**, no must-fix items.

Highlights beyond a code read:
- **Bidirectional oracle**: checked out the pre-fix `orient_capsule.py`, re-ran the new monkeypatch regression guard (`test_dedup_loop_never_calls_resolve_based_path_is_relative_to`) against it, and confirmed it **fails** (raises at the old dedup call site) — proving the guard is a real signal, not decoration.
- **Independent scaling reproduction**: measured the quadratic cost curve itself at a different N (80/150 candidates, not the PR's own 120/200/304/500) and got 1.92s→6.87s (3.57x time for 1.875x candidates, matching the N² prediction), with the fix flat at ~0.07-0.13s (~94x speedup at N=150) — corroborating the PR's numbers independently rather than trusting them.
- Verified the `deadline_hit` passthrough signature is keyword-only, so the new parameter cannot shift any positional args at the 3 other call sites (`repo_map.py:12755`, `session_store.py:1001`, `session_store.py:1177`), all of which omit it and stay byte-identical.
- Confirmed the lexical `.parts`-prefix swap is scoped to the *only* call site that doesn't need symlink-awareness — the other (security-relevant, containment-boundary) call sites of `_path_is_relative_to` (`repo_map.py:1178`, `1202`, `6273`, `10218`) are untouched.
- Confirmed the divergence between resolve()-based and lexical nesting is symlink-only and one-directional-safe: it can only *under*-dedup (a redundant nested entry, tolerated by the "first match wins" consumer), never over-dedup or mis-exclude.

Non-blocking notes for visibility: the real-workspace 1.24x-2.43x overshoot ratios were not replayed end-to-end (would require an unsafe multi-minute profiled run — the bug's own severity makes literal reproduction at that scale CPU-unsafe); the fix is instead justified by an independently-reproduced quadratic scaling law extrapolated to real-workspace candidate cardinality. `_reverse_importers`/`_reverse_import_distances` remain unbounded but were measured non-dominant (~3.4s vs 27.8s) and are left as a follow-up, consistent with this fix's scope (bound the identified dominant stage).

Pending CEO/maintainer decision on promoting from draft.

PR #682, opened 2026-07-20
feat(prepare): one-call edit-readiness CUJ — primary + blast-radius floor + validation + coordination hooks (CEO#5)

oimiragieo:
## Summary

- Adds `tg prepare REPO "task" [--json] [--deadline N] [--no-deadline] [--claim]` (CEO /goal #5,
  P1 flagship): a single edit-readiness CUJ call that replaces the
  orient -> search -> agent -> route-test -> callers -> evidence -> ledger tool loop. Returns, in
  one bounded call: primary target + confidence + `ask_user_before_editing`, a callers/blast-radius
  floor, validation commands, and claim/evidence coordination hooks.
- **Thin composition, no new ranking/scan logic** (structurally identical to `tg route-test` #223):
  `_build_prepare_payload` calls `build_agent_capsule` ONCE and reuses `primary_target`,
  `confidence`, `ask_user_before_editing`, `validation_plan`, and `validation_commands` verbatim.
- **The one new piece is the blast-radius floor** (`_build_prepare_blast_radius_floor`,
  `src/tensor_grep/cli/main.py`): the capsule's own call-site collector
  (`_collect_capsule_call_site_evidence`, `agent_capsule.py:514`) only collects when the query
  names the primary symbol AND its pre-cap confidence is >=0.75
  (`agent_capsule.py:536,546`) — a natural-language task query fails that gate and would
  otherwise leave the floor empty even though the SELECTED primary target has real callers.
  `prepare` reuses the capsule's own scan when it already ran
  (`blast_radius_floor.source = "capsule_call_site_evidence"`); otherwise it runs exactly ONE
  supplementary FS-backed scan via the literal-seed-rescue-equipped `build_symbol_blast_radius`
  keyed on the symbol the capsule actually selected, not the raw query text
  (`source = "supplementary_blast_radius"`).
- **Two load-bearing decisions, both deliberate:**
  1. **FS-backed floor via `build_symbol_blast_radius`, never `build_symbol_blast_radius_from_map`.**
     The `_from_map` sibling has no literal-seed rescue — a `no_match` against a possibly-truncated
     map reads as a false-complete "no callers" instead of "the scan window missed it." The rescue-
     equipped path is the one that recovers exactly the callers a pre-#108 cold scan would have
     found (see `_collect_capsule_call_site_evidence_from_map`'s own TRAP A docstring).
  2. **`--claim` is emit-args-only by default.** `coordination.claim` always carries a ready-to-run
     `tg ledger claim ...` command preview (`command`/`argv`), but `submitted` stays `false` and
     nothing is written to `.tensor-grep/ledger/` unless `--claim` is passed — a read-oriented
     command must not have a write side effect by default. `--claim` opts into an actual
     `ledger_store.submit_claim` call; a submit failure is caught and reported inline
     (`coordination.claim.error`), never raised, matching `submit_claim`'s own "never blocks"
     contract one level up.
- **Exit-2 honesty**: mirrors `route-test`/`agent` — the full JSON is printed FIRST, then `tg prepare`
  exits `2` (via the shared `_scan_incomplete` gate) when either the capsule build or the floor scan
  itself was truncated by `--deadline` (`partial: true`, `partial_reason: "deadline"`,
  `deadline_limit: {deadline_exceeded, capsule, blast_radius_floor}`). A plain `max_callers` OUTPUT
  cap on the floor (`possibly_incomplete: true`, no deadline cutoff) stays exit `0`, matching the
  existing output-cap-stays-`0` contract. Defaults its cold-only `--deadline` to
  `DEFAULT_AGENT_CLI_DEADLINE_SECONDS` (60s, the same constant `tg agent`/`tg route-test` use) rather
  than inventing a second default; `--no-deadline` disables the bound.
- **4 registration sites**, all updated: `KNOWN_COMMANDS` (`src/tensor_grep/cli/commands.py`), the
  Rust `Commands::Prepare` enum variant + `handle_python_passthrough` dispatch arm
  (`rust_core/src/main.rs`, mirrors `RouteTest` exactly), `PUBLIC_TOP_LEVEL_COMMANDS`
  (`tests/e2e/test_routing_parity.py`), and the `@app.command(name="prepare")` Typer function
  (`src/tensor_grep/cli/main.py`).
- Adds a `docs/CONTRACTS.md` §3 bullet documenting the composition, the floor's reuse-vs-supplementary
  contract, the claim default, and the exit-2 honesty tail (mirrors the `route-test` bullet directly
  above it).

## Rust build note

This box is CPU-safe-shared (never runs cargo/rustc/maturin here). The `Prepare` enum variant and
dispatch arm were written by hand, mirroring `RouteTest`'s syntax exactly, and verified by eye —
**not locally compiled**. CI's `test-rust-core` / `search-golden-parity` jobs build the real native
binary and are the authority on Rust correctness. One local test
(`tests/e2e/test_routing_parity.py::test_empty_invocation_fallback_help_matches_public_contract`)
fails locally ONLY because the pre-installed native `tg.exe` this box resolves from PATH
(`C:\Users\<user>\bin\tg.exe`, a managed install unrelated to this worktree — `rust_core/target/`
doesn't exist here) predates this change and cannot see the new `prepare` command in its
compiled-in clap fallback help. The sibling test
(`test_top_level_help_visible_commands_match_public_contract`) passes because with working Python
passthrough, native `--help` reflects Python's live `KNOWN_COMMANDS` — only the broken-Python
fallback path renders the stale compiled-in list. This resolves automatically once CI rebuilds the
binary from the updated `rust_core/src/main.rs`.

## Test plan

New TDD suite (real subprocess `python -m tensor_grep`, not CliRunner — mirrors
`tests/integration/test_agent_cold_deadline_tail_sla_220.py`'s harness):
`tests/integration/test_prepare_oneshot_cuj.py`, 8 tests, all passing:

- [x] Complete run: `primary_target`, `blast_radius_floor.callers_count>=1`, non-empty
      `validation_commands`, `coordination.claim.argv`, and an ABSENT honesty tail all in one call.
- [x] Validation commands are Python-shaped (pytest/py_compile, never npm/yarn) for a Python
      primary target.
- [x] Floor keyed on the SELECTED symbol, not the query: an NL query that names none of the
      fixture's 3 symbols verbatim still gets a real floor
      (`source: "supplementary_blast_radius"`, `callers_count>=1`) — proves the
      `agent_capsule.py:536` gate was bypassed, not silently left empty. Manually re-verified this
      exact scenario end-to-end against the real binary outside pytest (including a negative-control
      repro that omitted the caller file and correctly observed `callers_count: 0`, confirming the
      test isn't vacuously true).
- [x] `--claim` default is emit-args-only: `submitted: false` and `.tensor-grep/ledger/` is NOT
      created. Separate test: `--claim` creates the ledger dir, `submitted: true`,
      `result.claim.claim_id` present.
- [x] Tiny `--deadline 0.1` on a padded (~300-file) fixture exits `2` with `partial: true` +
      `partial_reason: "deadline"` + `deadline_limit.deadline_exceeded: true`, and the full JSON is
      still printed.
- [x] `--deadline 3` keeps wall-to-exit bounded (deadline + generous slack), outcome-agnostic
      (mirrors the #220 SLA test's own runner-speed-agnostic design).
- [x] `--no-deadline` smoke: runs to completion without a `partial` flag.
- [x] Registration RED-then-GREEN proof: temporarily renamed the `@app.command` decorator, confirmed
      `tests/unit/test_cli_bootstrap.py::test_typer_app_commands_match_source_of_truth` goes RED
      (`Extra items: 'prepare_TEMP_RED_PROOF'` / `'prepare'`), restored it, confirmed GREEN.

Regression suites run against the real worktree venv (`tensor_grep.__file__` verified to resolve
into this worktree, not the shared repo's stale installed copy):

- [x] `tests/e2e/test_routing_parity.py` — 52/53 pass; the 1 native-binary-dependent failure is the
      expected local-only staleness explained above.
- [x] `tests/unit/test_agent_capsule*.py` — 68/68 pass.
- [x] route-test suite (`-k "route_test or route-test"`) — 13/13 pass.
- [x] ledger suite (`-k ledger`) — 130/130 pass.
- [x] `tests/unit/test_main_cli_contracts.py` — 49/49 pass.
- [x] `tests/unit/test_cli_modes.py` (full file, 16.5k lines) — 523/523 pass.
- [x] docs-governance suite (`-k "docs_governance or contracts_docs or CONTRACTS"`) — 153 passed,
      1 skipped.
- [x] `tests/unit/test_skill_index_sync.py` — 4/4 pass.
- [x] `ruff check` — clean. `ruff format --check --preview` — clean (after one auto-format pass).
- [x] `mypy src/tensor_grep` (full package, strict mode) — clean, 81 source files, 0 errors (fixed
      one real `list[str]` vs `list[object]` invariance error caught on the first pass).

Not run: full `pytest -q` (multi-hour on this box for the whole tree) — ran the targeted suites
above instead, all touching either the new command's own surface or the shared registration/capsule
machinery it composes over.

---

**Not merging** — flagged for an independent Opus gate per the build brief (new command surface +
the 2 load-bearing decisions above: the FS-backed floor choice and the claim emit-only default).


oimiragieo:
Both gate nits closed, pushed (commit 8943a7f), still DRAFT.

**NIT-1** — added `orphan_repo` (isolated fixture, one symbol defined but never called anywhere) and `test_prepare_floor_reports_zero_callers_honestly` in `tests/integration/test_prepare_oneshot_cuj.py`. Asserts `blast_radius_floor.callers_count == 0`, `top_callers == []`, `source` is a real floor source (not `"error"` key present), and exit code 0. Manually cross-checked against the real binary outside pytest: `primary_target.symbol == "calculate_early_payment_discount"`, `blast_radius_floor == {callers_count: 0, source: "capsule_call_site_evidence", ...}`, no `error` key, exit 0 — confirmed genuine, not a fluke of the assertion. Pairs with the existing `test_prepare_floor_keyed_on_selected_symbol_not_query` (the `>=1` positive) as the bidirectional-oracle pair.

**NIT-2** — added the one-line(ish) comment at the `build_symbol_blast_radius` call in `_build_prepare_blast_radius_floor` (`src/tensor_grep/cli/main.py`), explaining `max_repo_files` is deliberately omitted: max recall within the shared deadline, not a second cap. Verified the mechanism before writing it — `_build_prepare_blast_radius_floor` only ever reads `radius_payload.get("partial")` back from the scan, never `radius_payload.get("scan_limit")`, so a file-count cap added later would truncate via `scan_limit` and be silently invisible in `tg prepare`'s output instead of surfacing through `partial`/exit-2.

N3/N4 left untouched per the gate's call.

Re-verified (real venv, `tensor_grep.__file__` resolves into the worktree):
- `tests/integration/test_prepare_oneshot_cuj.py` — 9/9 pass (8 previous + the new zero-caller control).
- `tests/e2e/test_routing_parity.py` — 52/53 pass; the 1 failure is the same pre-existing, unchanged `test_empty_invocation_fallback_help_matches_public_contract` (stale local native `tg.exe` on PATH predates the Rust `Prepare` enum addition from the first commit — CI rebuilds the binary).
- `ruff check` — clean. `ruff format --check --preview` — clean.
- `uv run mypy src/tensor_grep` (full package, strict) — clean, 81 files, 0 errors.

Not un-drafting — leaving that to you per instructions.

PR #691, opened 2026-07-21
perf(agent): bound the post-deadline assembly-tail residual (#222 hard-SLA)

oimiragieo:
## Summary

Pins and fixes the residual super-linear stage behind issue #222 (`tg agent`'s
post-deadline assembly-tail overshoot on a mega-repo) that #669/#671 left
un-gated: `_build_context_pack_from_map`'s reverse-import-**graph** construction.

- **`_reverse_import_distances`** (a 3-depth BFS): had **no deadline check at
  all**. Cost is O(depth x files x frontier), and frontier grows with any real
  import fan-in. Measured **~n^2.2** with file count on a direct, non-subprocess
  probe: 0.99s @ 2,000 files -> 13.6s @ 6,000 (13.8x) -> 60.3s @ 12,000 (61x for
  6x files). At 12,000 files this was **84%** of `_build_context_pack_from_map`'s
  own total cost (60.3s of 71.5s).
- **`_reverse_importers`**: same un-gated call block, cheaper/near-linear but
  still unbounded.
- A **third**, independently-discovered consumer: found via an OLD-vs-NEW
  re-profile of the fix above — a direct `_import_graph_bonus` loop later in the
  same function dominated the *post-fix* profile (93s of 102s @ 12,000 files)
  because a query term that fuzzy-matches many files' import strings can pull
  hundreds-to-thousands of files into `dependency_seed_files`.
- **`build_symbol_blast_radius_from_map`**: the identical un-gated pattern in
  its own direct reverse-import-graph derivation (reached from `tg agent`'s
  cold path via the call-site-evidence blast-radius call).
- A **separate, smaller, count-bounded** (not super-linear) gap found while
  verifying end-to-end via the real CLI: `_build_edit_plan_seed`'s validation-
  runner detection falls back to a fresh, un-deadlined `_iter_repo_files` walk
  in 4 places, even though `_iter_repo_files` already supports
  `deadline_monotonic`/`deadline_hit` and the enclosing functions already
  accept both.

Fix shape mirrors the exact established pattern (#669/#671/pagerank/vendored-
subtree-detection): a per-item deadline check inside the expensive inner loop
(not just an outer-loop check — the BFS's outer loop only runs 3 times
regardless of repo size, so it can't bound anything). On expiry each function
returns the **partial** result already accumulated — never discarded, matching
how every downstream consumer already treats a missing entry as "no signal for
this file."

**OLD-vs-NEW, real CLI, 6,000-file synthetic repo, `--deadline 3.0`, cold
filesystem cache:** wall-to-exit **26.6s -> 9.5s**.

## Why this magnitude is cache-sensitive (documented in the test docstrings)

The dominant per-item cost is Windows `Path.resolve()`/syscall latency (same
class #671 documented), so wall-clock severity swings a lot with filesystem-
cache warmth — a warm dev-box re-run of the same fixture shows both pre- and
post-fix collapsing toward each other (~6-9s). This is real, not test
flakiness. The **primary, cache-immune regression guard is the 11 new
deterministic unit tests** (monkeypatched clock, no real timing dependency);
the 1 new integration test uses a generous absolute ceiling for the same
reason every sibling #220/#222 SLA test does.

## Test plan

- [x] 11 new deterministic unit tests (`tests/unit/test_agent_reverse_import_graph_deadline_222.py`) — inner-loop-check-fires, partial-not-discarded, no-pressure-byte-identical, for all 3 fixed loops + `build_symbol_blast_radius_from_map` + the 4 validation-runner fallback threads.
- [x] 1 new real-subprocess integration test (`tests/integration/test_agent_reverse_import_graph_scale_sla_222.py`) proving the CLI-level SLA + honesty contract (`partial`/`deadline_limit`) on a 6,000-file synthetic repo.
- [x] Verified the new integration test is a real regression guard (red without the fix, green with it) via a stash/pop OLD-vs-NEW.
- [x] `mypy src/tensor_grep` clean (81 files).
- [x] `ruff check` clean, `ruff format --check --preview` clean.
- [x] Targeted suite green: 452 unit + 11 integration tests spanning agent/deadline/blast-radius/validation/inventory/file-deps/mcp-find/session-cli. One pre-existing test double (`test_validation_commands.py`'s `fake_iter_repo_files` stub) needed its signature widened to accept the two new optional kwargs — fixed in this PR.
- [x] Did NOT touch Rust/native code, did not run cargo/maturin, did not run `tests/e2e/test_routing_parity.py` per the assignment's hard constraints.

### DIFF
--- a/src/tensor_grep/cli/repo_map.py
+++ b/src/tensor_grep/cli/repo_map.py
@@ -3865,21 +3865,36 @@ def _relevant_tests_for_symbol(
             str(entry["file"]): [str(item) for item in entry["imports"]]
             for entry in repo_map.get("imports", [])
         }
+        # #691 gate NIT-1 (#222 residual, still-reachable cold path): this function already
+        # declares `deadline_monotonic`/`deadline_hit` (used by the direct_definition_tests loop
+        # below) but left THESE three whole-repo graph calls un-gated -- the identical ~n^2.2 BFS
+        # `_reverse_import_distances` bounded elsewhere in this module. `build_symbol_callers_
+        # from_map` reaches this exact block with BOTH `caller_files=` (non-None, so this branch
+        # runs) AND a real `deadline_monotonic` (the `tg callers --deadline SYMBOL` budget), so a
+        # high-fan-in symbol could still blow the whole-repo BFS/reverse-index cost this PR exists
+        # to bound. Thread the two kwargs already in scope, same per-item-inside-the-expensive-
+        # loop shape as every other call site this PR fixed.
         reverse_importers = _reverse_importers(
             all_files,
             imports_by_file,
+            deadline_monotonic=deadline_monotonic,
+            deadline_hit=deadline_hit,
             _profiling_collector=_profiling_collector,
         )
         file_distances = _reverse_import_distances(
             source_files,
             all_files,
             imports_by_file,
+            deadline_monotonic=deadline_monotonic,
+            deadline_hit=deadline_hit,
             _profiling_collector=_profiling_collector,
         )
         graph_scores = _personalized_reverse_import_pagerank(
             source_files,
             all_files,
             reverse_importers,
+            deadline_monotonic=deadline_monotonic,
+            deadline_hit=deadline_hit,
             _profiling_collector=_profiling_collector,
         )
         file_scores: dict[str, int] = {}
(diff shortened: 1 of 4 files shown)


oimiragieo:
## Gate follow-up (independent Opus review: SHIP-WITH-NITS)

Folded in both nits on this same branch (commit 21bf2a6):

**NIT-1 (core-value, required before merge):** `_relevant_tests_for_symbol` (`repo_map.py:3825`) already declared `deadline_monotonic`/`deadline_hit` in its own signature but ran `_reverse_importers`/`_reverse_import_distances`/`_personalized_reverse_import_pagerank` un-gated inside its `if caller_files:` block. `build_symbol_callers_from_map` reaches this exact block with both a non-None `caller_files` and a real `deadline_monotonic` (the actual `tg callers --deadline SYMBOL` budget), so a high-fan-in symbol on that cold path — and the agent capsule's caller-evidence path, which calls into the same builder — could still run the unbounded whole-repo BFS this PR exists to eliminate. Threaded the two kwargs already in scope into all three calls (`repo_map.py:3868/3873/3879` at authoring time), identical shape to every other site this PR fixed. Corrected `_reverse_import_distances`'s own docstring, which claimed the context-pack call was "the one un-gated... consumer" — no longer accurate with a second reachable call site.

**NIT-2 (symmetry, lower severity):** `build_file_importers_from_map` (`tg importers`, `repo_map.py:16247`) had `deadline_monotonic` in scope and its own confirm-edges loop already honored it, but left its `_reverse_importers` call (`repo_map.py:16307`) un-threaded. Folded into the same `deadline_hit` local the confirm-edges loop already sets.

**Tests:** 6 new deterministic unit tests added to `tests/unit/test_agent_reverse_import_graph_deadline_222.py` (now 17 total in that file) — wiring + mechanism + reachability + no-pressure regression guards for NIT-1, plus forwarding + partial-folding guards for NIT-2. Verified via stash/pop that all 4 mechanism/wiring tests fail without this diff's source changes and pass with them.

**Gates:** `mypy src/tensor_grep` clean (81 files), `ruff check` clean, `ruff format --check --preview` clean. Targeted suite green: 649 unit + 11 integration tests.

PR #693, opened 2026-07-21
fix(agent-capsule): down-weight thin CLI-dispatcher wrappers in primary-target ranking (#250)

oimiragieo:
## Summary

`tg prepare src/tensor_grep "fix the ledger claim TTL logic"` resolved `primary_target` to
`cli/main.py`'s `ledger_claim` -- a `@ledger_app.command("claim")` Typer dispatcher that is a
single call-through to `ledger_store.submit_claim(...)` -- instead of the real implementation in
`cli/ledger_store.py` (which owns `_DEFAULT_TTL_SECONDS`/`_TTL_ENV`/`_configured_ttl_seconds` and
`ClaimRecord.ttl_seconds`). Root cause: `ledger_claim`'s name is a literal concatenation of the
query's two substantive words ("ledger" + "claim"), so `repo_map._score_symbol`'s
`covered_query_match` bonus (embedded +2 in the score, +24 at the file-score level, plus top-rank
bucket precedence in `_symbol_rank_key`) let the dispatcher outrank the implementation even though
`ledger_store.py` scored higher on every other signal (path, import, source, graph-centrality).

## Fix

A conservative, structural down-weight applied **after** ranking -- not a change to the scorer
itself (which stays untouched, so nothing else shifts):

- `repo_map._thin_cli_dispatcher_call_targets` (`repo_map.py:6938`, gated by
  `_is_cli_command_module_path` at `repo_map.py:6921`) reuses the already-cached AST parse of the
  primary target's own file to check whether the symbol is **(a)** decorated as a Typer/Click
  command (`.command(...)`) and **(b)** whether its body calls a specific alternative candidate
  directly. No new repo-wide scan -- one bounded re-check of a symbol already selected as primary.
- Only when **both** hold does `agent_capsule._prefer_implementation_over_cli_dispatcher_helper`
  (`agent_capsule.py:401`, via `_cli_dispatcher_implementation_candidate` at
  `agent_capsule.py:340`) swap it with that alternative -- mirroring the existing
  `_prefer_implementation_over_marker_helper` pattern already in this file. The dispatcher is
  demoted to an alternative, never dropped, so downstream tie/ambiguity signals still apply.
- Wired into both capsule-shaping call sites (`agent_capsule.py:2100`, `:2721`), right after the
  marker-helper swap, so `tg prepare`, `tg agent`, and `tg edit-plan` all see the fix identically.

The crux guard: a bare "primary lives under `cli/`" is deliberately **not** enough --
`cli/main.py` is the correct target for plenty of tasks (e.g. "add a --flag to tg search", where
the flag registration itself lives in that Typer command's signature). The swap only fires when
the primary is **provably** a thin pass-through (decorator + call-through both present), never on
lexical score or file location alone.

## Validation -- the accuracy-gate numbers (`tests/eval/test_agent_accuracy.py`)

| Run | Golden set | Result |
|---|---|---|
| BEFORE (baseline, no fix) | original 15 tasks | **15/15** |
| AFTER (fix applied) | same 15 tasks | **15/15** -- byte-identical `primary_target.file` per task, including "add a new --flag to tg search" still resolving to `cli/bootstrap.py` |
| AFTER (fix applied) | 16 tasks ("fix the ledger claim TTL logic" added back) | **16/16** -- the previously-dropped task now resolves to `cli/ledger_store.py` |

The 16th task was dropped from the original golden set (PR #690) specifically because of this bug;
it's added back here now that it hits. Floor bumped `12 -> 13` to preserve the documented "3 tasks
of slack" policy against the new 16-task baseline.

## Test plan

- [x] `ruff check .` -- clean
- [x] `ruff format --check --preview .` -- clean
- [x] `mypy src/tensor_grep` -- clean (81 files)
- [x] New unit tests: 4 in `tests/unit/test_repo_map_targets.py` (AST-level detector: command-call-through,
      no-decorator, outside-cli, symbol-missing) + 6 in `tests/unit/test_agent_capsule_hardcases.py`
      (swap helper, including the crux guard that a genuinely-implementing `cli/main.py` primary is
      NOT demoted, plus a full synthetic end-to-end capsule repro) -- all passing
- [x] `tests/integration/test_prepare_oneshot_cuj.py` -- 9/9 passing
- [x] `tests/eval -m eval -v -s` (the accuracy gate) -- before AND after, see table above
- [x] Broader regression sweep (~900 tests across agent_capsule/repo_map/codemap/edit_plan/context/cli_modes):
      4 pre-existing failures in `test_cli_modes.py` (terminal-width-dependent `--help` rendering)
      reproduce **identically** on origin/main with this diff stashed out -- confirmed unrelated to
      this change (same failures, same messages, with or without the fix)

## Gate note

LOAD-BEARING: `primary_target` ranking is core capsule value (per `AGENTS.md`/`CLAUDE.md`
change-control discipline). This needs an **independent Opus gate** before merge to verify it does
not regress the golden set or shift other `primary_target`s -- not self-gated in this PR.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

### DIFF
--- a/src/tensor_grep/cli/agent_capsule.py
+++ b/src/tensor_grep/cli/agent_capsule.py
@@ -337,6 +337,84 @@ def _prefer_implementation_over_marker_helper(
     return implementation, demoted
 
 
+def _cli_dispatcher_implementation_candidate(
+    primary_target: dict[str, Any],
+    alternatives: list[dict[str, Any]],
+) -> dict[str, Any] | None:
+    """Task #250: return the alternative a thin CLI-dispatcher primary target provably calls
+    through to, or ``None`` if the primary is not a provable pass-through.
+
+    Conservative by design: a bare "primary lives under cli/" is NOT enough on its own --
+    `cli/main.py` is the CORRECT target for plenty of tasks (e.g. "add a --flag to tg search",
+    where the flag registration itself lives in that Typer command's own signature). This only
+    fires when BOTH:
+
+      1. the primary symbol is decorated as a Typer/Click command (``@x.command(...)``) -- the
+         structural signature of a dispatcher, not a guess from its name or lexical score; and
+      2. the primary symbol's OWN body (``_thin_cli_dispatcher_call_targets`` -- a bounded,
+         already-selected span, not a new scan) contains a direct call to a SPECIFIC alternative
+         candidate's symbol, defined in a DIFFERENT file -- i.e. the dispatcher hands off to that
+         exact implementation.
+
+    Real repro (task #250): ``tg prepare`` resolving "fix the ledger claim TTL logic" to
+    ``cli/main.py``'s ``ledger_claim`` (a ``@ledger_app.command("claim")`` dispatcher whose body
+    is a single call to ``ledger_store.submit_claim(...)``) instead of the real implementation in
+    ``cli/ledger_store.py``, purely because ``ledger_claim`` happens to lexically match both
+    query words at once. This helper recognizes that shape and prefers the callee.
+    """
+    file_path = str(primary_target.get("file") or "")
+    symbol_name = str(primary_target.get("symbol") or "")
+    if not file_path or not symbol_name:
+        return None
+    if str(primary_target.get("kind") or "") not in {"function", "method"}:
+        return None
+
+    candidates = [
+        alternative
+        for alternative in alternatives
+        if str(alternative.get("symbol") or "")
+        and str(alternative.get("file") or "") not in ("", file_path)
+        and str(alternative.get("kind") or "") in {"function", "method"}
+    ]
+    if not candidates:
+        return None
+
+    line = primary_target.get("line")
+    expected_line: int | None = None
+    if isinstance(line, int):
+        expected_line = line
+    elif isinstance(line, str) and line.isdigit():
+        expected_line = int(line)
+    called_names = repo_map._thin_cli_dispatcher_call_targets(
+        file_path, symbol_name, expected_line=expected_line
+    )
+    if not called_names:
+        return None
+
+    matched = [candidate for candidate in candidates if str(candidate["symbol"]) in called_names]
+    if not matched:
+        return None
+    matched.sort(key=lambda candidate: _numeric_confidence(candidate.get("confidence"), 0.0))
+    return matched[-1]
+
+
+def _prefer_implementation_over_cli_dispatcher_helper(
+    primary_target: dict[str, Any],
+    alternatives: list[dict[str, Any]],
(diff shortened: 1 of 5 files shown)


oimiragieo:
## NIT-1 addressed (structural thinness gate)

The independent Opus gate flagged that `_thin_cli_dispatcher_call_targets` had no real
"thinness" check -- `search_command` (cli/main.py:6648, the actual ~1500-line `tg search`
implementation: 89 top-level statements / 104 distinct callees, docstring excluded) is
ALSO a `.command`-decorated function, so the swap's safety was an emergent property of
which names happen to rank as alternatives, not a structural guarantee.

**Fix** (commit `30e3dac`): two calibrated thresholds, both checked in
`_thin_cli_dispatcher_call_targets` (`repo_map.py`) right after the decorator check:

| Threshold | Value | Rationale |
|---|---|---|
| `_THIN_DISPATCHER_MAX_BODY_STATEMENTS` | 20 | top-level body statements, docstring excluded |
| `_THIN_DISPATCHER_MAX_CALL_TARGETS` | 20 | distinct callee names (reuses the already-built `called_names` set -- no extra AST walk) |

Calibrated against the two known genuine dispatchers -- `ledger_claim` (10 / 14) and
`ledger_release` (6 / 10) -- vs. the one known fat command, `search_command` (89 / 104).
Both real dispatchers sit at/under 15 on each axis; `search_command` sits 6-10x over. 20
leaves ~2x headroom above the largest observed genuine dispatcher while staying far below
any real implementation function measured in this repo. Docstrings in both
`repo_map.py`/`agent_capsule.py` corrected to state the true 3-part precondition
(decorator AND structural thinness AND call-through).

**New regression tests** (the exact fragility flagged, verified to FAIL when the two
threshold checks are temporarily disabled -- not vacuous):
- `test_thin_cli_dispatcher_call_targets_none_for_fat_command`
- `test_thin_cli_dispatcher_call_targets_none_for_many_distinct_callees_even_if_few_statements`
- `test_prefer_implementation_no_swap_for_fat_command_even_if_a_callee_matches` (end-to-end:
  a synthetic 50-call command with a deliberately-matching callee name is correctly NOT demoted)

**Delta-verify**: `tests/eval -m eval -v -s` stays **16/16**, byte-identical to before this
commit -- the genuinely thin `ledger_claim` (10/14, both under 20) still triggers the swap,
"fix the ledger claim TTL logic" still resolves to `cli/ledger_store.py`, and "add a new
--flag to tg search" is unaffected. ruff/mypy/45 unit tests/9 integration tests/380-test
regression sweep all green.

NIT-2 (accuracy-gate floor slack) intentionally left untouched, tracked separately.

PR #698, opened 2026-07-21
fix(search): detect binary content in chunk-parallel native search

oimiragieo:
---
_Independent Opus gate (SHIP-WITH-NITS): parity verified against the pinned grep-searcher **0.1.16**. Residual documented in-code — beyond the guaranteed 64 KiB floor the serial path opportunistically scans matched lines; this fix does not, an under-flag-only, safe-direction gap that is a strict Pareto improvement over the pre-fix raw-match behavior._

### DIFF
--- a/rust_core/src/native_search.rs
+++ b/rust_core/src/native_search.rs
@@ -28,6 +28,15 @@ const JSON_OUTPUT_VERSION: u32 = 1;
 const LARGE_FILE_CHUNK_THRESHOLD_BYTES: usize = 50 * 1024 * 1024;
 const STREAMING_OUTPUT_FLUSH_BYTES: usize = 64 * 1024;
 const STREAMING_OUTPUT_FLUSH_BYTES_DEBUG: usize = 8 * 1024;
+/// Mirrors `grep_searcher::line_buffer::DEFAULT_BUFFER_CAPACITY` (64 KiB): the fixed-size prefix
+/// that `grep-searcher`'s `BinaryDetection::quit` guarantees to scan for the binary byte when
+/// searching mmap-backed content (see that type's doc comment in the `grep-searcher` crate --
+/// "only a fixed sized region at the beginning of the contents are detected for binary data").
+/// The serial (non-chunked) search path relies on exactly this guaranteed floor when a whole file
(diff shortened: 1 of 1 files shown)


oimiragieo:
## Independent gate verdict: SAFE TO UN-DRAFT

An independent review (fresh-eyes, re-derived the `grep-searcher` binary-detection semantics from the pinned `0.1.17` source rather than trusting this PR's description) confirmed:

- **Scope clean**: only `rust_core/src/native_search.rs` changed vs `main`.
- **grep-searcher parity verified against the real crate source**: `BINARY_DETECTION_PREFIX_BYTES = 64*1024` matches `line_buffer::DEFAULT_BUFFER_CAPACITY`; the guaranteed-floor scan range (`Range::new(0, min(slice.len(), 64KiB))` in `searcher/glue.rs`) matches what `detect_binary_prefix` reproduces; the NUL byte and `config.text` opt-out match `build_searcher`'s `BinaryDetection::quit(b'\x00')` / `BinaryDetection::none()`.
- **Correctness sound**: early return precedes both parallel scans; result shape (`matches`, `match_count`, `binary_detected`, `binary_match_detected` via the same `binary_file_matches_pattern` helper, `binary_byte_offset`) is byte-identical to what the serial leaf functions return for the same binary content; text-file behavior is a strict no-op; no borrow/panic risk.
- **Tests valid**: the multi-chunk sanity check is self-protecting and the `binary_detected == true` assertion catches the regression against the old hardcoded-`false` code.

Two non-blocking nits were raised and have both been fixed in a follow-up commit (`116dffc`):
1. The binary test fixture now embeds the search pattern in its surrounding text, so the `match_count`/`matches.is_empty()` assertions actually discriminate old-buggy vs fixed behavior (previously they'd have passed either way, since nothing matched regardless).
2. The inline comment describing the fix's fidelity to the serial path now says "guaranteed detection floor" rather than "exactly," and explicitly notes the one documented, conservative gap (this fix doesn't reproduce grep_searcher's secondary opportunistic scan of matched/context-line bytes beyond the 64 KiB floor — a gap that can only under-flag, never over-flag, relative to the serial path).

`cargo fmt --check` verified clean after both fixes. Marking ready for review.

PR #702, opened 2026-07-22
fix(search): close #105 gap -- front-door raw-rg-passthrough had no walk-ceiling guard

oimiragieo:
## Verdict

**Classification: (a) -- a REAL residual gap on an unguarded shape**, reproduced natively (Windows, no WSL involved). Hypothesis (ii) (silent empty-exit-0 on timeout) was investigated and **falsified** -- the existing timeout path is already fail-closed. Hypothesis (c) (pure WSL artifact) is **not the explanation** for this specific gap -- it reproduces on any OS whenever the shape below is hit, though WSL `/mnt/c` amplification (documented separately, `tensor-grep-wsl-mnt-c-latency-artifact-2026-07-11`) could still worsen wall-clock on top of it.

## Root cause

`bootstrap.py`'s front door has **two independent `rg`-invocation paths**:

1. Delegate to the **native `tg` binary**, which internally shells out to `rg` via `rust_core/src/rg_passthrough.rs` -- protected by `check_implicit_walk_ceiling` (audit #100/#105/#109), verified via direct invocation to refuse a 2420-file implicit-path root in **~34ms**.
2. `_run_rg_passthrough` (`bootstrap.py:1007`) -- a **raw `rg` subprocess spawn straight from Python**, bounded ONLY by a wall-clock timeout (`TG_RG_TIMEOUT_SECONDS`/`TG_SIDECAR_TIMEOUT_MS`, default 60s). **No walk-ceiling probe at all.**

Bootstrap's dispatch logic **prefers path 2 for the common, flag-less `tg search PATTERN` shape**: `_can_delegate_to_native_tg_search` (`bootstrap.py:443`) only returns `True` when a "supported trigger" flag is present (`--cpu`/`--force-cpu`/`--json`/`--ndjson`/`--gpu-device-ids`), and `TG_RUST_FIRST_SEARCH` (the opt-in that would prefer native first) is **off by default**. So a bare, unscoped `tg search PATTERN` -- the exact shape in the dogfood report -- almost never reaches the native binary's own ceiling guard; it goes straight to the unguarded raw-`rg` path whenever `rg` resolves.

This is compounded by how the native binary is actually distributed: a plain `pip install tensor-grep` does **not** install a standalone native `tg` binary (only `[project.scripts] tg = "tensor_grep.cli.bootstrap:main_entry"` -- a Python entry point; the compiled binary is a *separate* managed install via curl/npm/Homebrew/winget downloading a GitHub release asset into `~/.tensor-grep/bin/`). So `resolve_native_tg_binary()` legitimately returns `None` for a large class of real installs -- very plausible for a pip-based WSL dogfood environment -- which makes path 2 (raw rg-passthrough, unguarded) the *default* experience, not an edge case.

## Classification matrix (shape x guard x outcome, file:line)

All measured on a synthetic 2420-file single-project root (one `.git` marker only, no vendored dir name, no marked siblings -- deliberately slips past every existing guard) on native Windows NTFS, via `python -m tensor_grep search PATTERN` (the real bootstrap front door, not CliRunner).

| # | Shape | Guard that fires (file:line) | Outcome before fix | Outcome after fix |
|---|---|---|---|---|
| 1 | Marked-workspace root, unscoped | `_search_paths_include_workspace_root` (`bootstrap.py:673`, item #154) | exit 2, fast refuse | unchanged |
| 2 | Vendored top-level dir, unscoped | `_search_paths_include_vendored_root` (`bootstrap.py:735`) | exit 2, fast refuse | unchanged |
| 3 | **Plain large single-project root, unscoped, native binary resolvable** | none (native never invoked -- no supported-trigger flag) | ran raw `rg`, **0.29-3.16s, no refuse** | **exit 2, ~0.35s, fast refuse** |
| 4 | **Plain large single-project root, unscoped, native unresolvable, `rg` resolvable** | none | ran raw `rg`, **no refuse** | **exit 2, ~0.35-0.87s, fast refuse** |
| 5 | Same, native+rg both unresolvable | `_should_refuse_unbounded_large_root_scan` (`main.py`, Bug #88, already shipped) | exit 2, fast refuse (full CLI) | unchanged (still the same path) |
| 6 | Same root, short `TG_RG_TIMEOUT_SECONDS=0.001` | `_streaming_passthrough_returncode` timeout handler (`bootstrap.py:900`) | **exit 124 + honest stderr** (already fail-closed) | now refused before timeout even matters (~0.38s) |
| 7 | Same root, **explicit path** (`search PATTERN .`) | `paths_defaulted` gate (Trap #3) | ran normally, found match, ~0.29s | **unchanged** -- 0.26-0.30s across 3 repeats, zero measurable regression |
| 8 | Small tree (40 files), unscoped | ceiling probe returns False | ran normally | unchanged |
| 9 | Large tree, unscoped, explicit `--max-depth` reaching target | `max_depth` unconditional-bound gate | ran normally | unchanged |
| 10 | Native binary present + `--json`/`--cpu` trigger, unscoped, oversized | `check_native_implicit_walk_ceiling` (Rust, already shipped) | exit 2, ~34ms (direct) | unchanged |

Row 3-4 are the confirmed gap; rows 1-2, 5-6, 8-10 are non-regression / already-correct baselines re-verified during this investigation.

## Fix

- `src/tensor_grep/io/directory_scanner.py`: new `IMPLICIT_SEARCH_WALK_FILE_CEILING = 1500` -- single source of truth (same pattern as `UNBOUNDED_VENDORED_ROOT_DIR_NAMES` / `BROAD_WORKSPACE_PROJECT_MARKERS`).
- `src/tensor_grep/cli/main.py`: `_LARGE_ROOT_SCAN_FILE_CEILING` now imports that constant instead of a separate hardcoded `1500` literal (prevents future drift between the two Python-side guards).
- `src/tensor_grep/cli/bootstrap.py`: new `_search_paths_include_oversized_implicit_root(paths, search_args)` -- a bounded walk (never more than ceiling+1 entries) via `DirectoryScanner`, mirroring `cli/main.py`'s existing `_implicit_glob_search_walk_exceeds_ceiling` (Bug #88). Wired into `_search_args_include_unbounded_broad_scan`, gated on `paths_defaulted` (implicit path only -- an explicit PATH still runs uninhibited, matching the existing Trap #3 contract). Also widens the probe to `no_ignore=True` when any `--no-ignore*` flag is present in the raw args, so it never under-counts relative to the real walk.
- When the guard fires, bootstrap defers to `_run_full_cli()`, which already owns the actual refusal message + exit 2 via `cli/main.py`'s pre-existing `_should_refuse_unbounded_large_root_scan` (verified working end-to-end before this PR -- this fix is about *reaching* that guard, not building a new one).

## Non-regression proof (measured, not assumed)

- **Scoped search (explicit path) on the identical oversized tree: unaffected.** The probe is gated on `paths_defaulted`, so it never runs for a scoped search -- 3 repeated timings post-fix: 0.301s / 0.284s / 0.287s, matching pre-fix baseline timings (0.29-0.35s) with no measurable delta.
- Small tree, unscoped: still runs normally (no over-refusal).
- `--max-depth` on an oversized implicit root: still runs normally (existing escape hatch preserved).
- `--allow-broad-generated-scan` opt-in: still bypasses the new guard.
- 85/86 existing `test_cli_bootstrap.py` tests pass (1 skip: native-binary-required lock test, no in-tree build in this environment). 522/523 `test_cli_modes.py` tests pass (1 pre-existing failure, `test_cli_search_warns_when_gpu_device_id_out_of_local_inventory`, confirmed failing identically on unmodified `main` via `git stash` -- unrelated to this change, GPU device-inventory environment issue).
- 7 new tests added: constant-parity, probe-direct (over/under ceiling, `--no-ignore` widening), full `main_entry()` dispatch-level (oversized implicit refuses via full CLI / oversized explicit still fast-paths), and `_search_args_include_unbounded_broad_scan`-level.
- `ruff check` clean, `ruff format --check --preview` clean, `mypy` clean on all touched files.

## What did NOT change

- No new CLI flag or command (no registration-site changes needed).
- The fail-closed timeout contract (`_streaming_passthrough_returncode`, exit 124 + honest stderr) was already correct -- untouched, just re-verified.
- `prepare`/`doctor`/find-hint copy and the ledger were not touched (out of this seam).

## Flag for review

**Front-door change** (`bootstrap.py` dispatch logic) -- flagging for an independent Opus gate per house discipline.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

### DIFF
--- a/src/tensor_grep/cli/bootstrap.py
+++ b/src/tensor_grep/cli/bootstrap.py
@@ -751,6 +751,56 @@ def _search_paths_include_vendored_root(paths: list[str]) -> bool:
     return False
 
 
+# Item #105 (bootstrap raw-rg-passthrough gap, CEO dogfood v1.92.x directive): neither
+# `_search_paths_include_workspace_root` above (needs >=3/>=8 independently-MARKED sibling
+# dirs) nor `_search_paths_include_vendored_root` (needs a top-level vendored dir NAME) catches
+# a plain, single, large repo root -- e.g. a flat monorepo `src/` with thousands of files, no
+# vendored subdir, and no marked siblings. That shape sailed straight into `_run_rg_passthrough`
+# (a raw `rg` subprocess spawn bounded ONLY by a wall-clock timeout --
+# `TG_RG_TIMEOUT_SECONDS`/`TG_SIDECAR_TIMEOUT_MS`, no proactive refusal) because bootstrap's fast
+# path never delegates a bare, flag-less search to the native binary by default --
+# `_can_delegate_to_native_tg_search` requires a "supported trigger" flag (`--cpu`/
+# `--force-cpu`/`--json`/`--ndjson`/`--gpu-device-ids`) and `TG_RUST_FIRST_SEARCH` is off by
+# default -- so the native binary's OWN `check_implicit_walk_ceiling`
+# (`rust_core/src/rg_passthrough.rs`, audit #100/#105/#109, verified via direct invocation to
+# refuse in ~34ms) never gets a chance to run for the common bare `tg search PATTERN` shape.
+# `cli/main.py`'s full-CLI path already refuses this shape correctly (Bug #88 fix,
+# `_should_refuse_unbounded_large_root_scan`) -- this mirror sends the SAME unscoped search
+# there instead of into an unguarded raw `rg` spawn, matching the pattern already used above for
+# the workspace-root/vendored-root guards (both of which bypass cli/main.py's Python guards
+# entirely without a front-door mirror).
+#
+# Deliberately WALKS (bounded to ceiling+1 entries, never a full-tree enumeration) rather than a
+# one-level `iterdir()` probe like the vendored/workspace guards above -- there is no shallow
+# signal for "this tree is huge"; the walk itself is the only honest measurement, same approach
+# as cli/main.py's `_implicit_glob_search_walk_exceeds_ceiling`. Mirrors `--hidden`/
+# `--no-ignore*` flags into the probe config (conservatively: ANY no-ignore-family flag widens
+# the probe to `no_ignore=True`, the most inclusive walk `DirectoryScanner` supports) so this
+# probe does not UNDER-count relative to the real invocation and let an oversized
+# `--hidden`/`--no-ignore` walk slip through as "under ceiling".
+def _search_paths_include_oversized_implicit_root(paths: list[str], search_args: list[str]) -> bool:
+    from tensor_grep.core.config import SearchConfig
+    from tensor_grep.io.directory_scanner import (
+        IMPLICIT_SEARCH_WALK_FILE_CEILING,
+        DirectoryScanner,
+    )
+
+    probe_config = SearchConfig(
+        hidden=any(arg in _SEARCH_HIDDEN_FLAGS for arg in search_args),
+        no_ignore=any(arg in _SEARCH_NO_IGNORE_FLAGS for arg in search_args),
+    )
+    count = 0
+    for raw_path in paths:
+        if not raw_path or raw_path == "-" or raw_path.startswith("-"):
+            continue
+        scanner = DirectoryScanner(probe_config)
+        for _ in scanner.walk(raw_path):
+            count += 1
+            if count > IMPLICIT_SEARCH_WALK_FILE_CEILING:
+                return True
+    return False
+
+
 def _search_args_include_unbounded_broad_scan(search_args: list[str]) -> bool:
     if "--allow-broad-generated-scan" in search_args:
         return False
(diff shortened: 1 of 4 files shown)


oimiragieo:
Independent Opus gate: **SHIP-WITH-NITS — no false-refused shape, no contract break** (full shape×door matrix verified empirically at the PR head; all three doors coherent at 1500; bootstrap never refuses itself — it routes to the full CLI which independently re-decides via the pre-existing P0-1 guard). Non-blocking follow-ups banked: (1) mirror the sibling's specific `no_ignore_vcs`/`no_ignore_files` fields instead of collapsing the no-ignore family (latency-only over-count on `--no-ignore-dot`-class shapes); (2) pin the bounded-probe cost with a regression test (empirically flat 72-77ms at 3.5k/9.5k files); (3) add a test-author comment on bare-search cwd-sensitivity. Reviewer context (deliberate, not a defect): this extends the existing 1500-file refuse policy to the last unguarded door, so a bare `tg search PAT` on a >1500-file root now fast-refuses with remediation by default — the exact behavior the v1.92.1 dogfood requested, escapable via PATH scoping or `--allow-broad-generated-scan`.

PR #704, opened 2026-07-22
fix(gpu): detect a bare-named WSL front-door shim as cross-domain (#171 GPU-P0-1 follow-up)

oimiragieo:
## Verdict

**(b) — a real, reproduced sibling path the #184/#594 (`GPU-P0-1`) fix missed.**
Not a recurrence of the original bug, and not honest-but-misread behavior: `wslpath`
translation genuinely fixes it, live-verified end to end on the reporting box.

## Probe-path map

Every place a filesystem path crosses from a WSL Python process to a resolved native `tg`
binary, and whether it goes through translation:

| Call site | Resolves binary via | Cross-domain gate | Status before this PR |
| --- | --- | --- | --- |
| `tg doctor` GPU probe — `_doctor_gpu_search_runtime_probe` (`src/tensor_grep/cli/main.py:2882-2925`) | `resolve_native_tg_binary()` (always a resolved `Path`, never a bare string) | `is_cross_domain_native_binary()` (`runtime_paths.py:2907`) | **Missed** the bare-named shim case |
| Agent GPU evidence — `_agent_gpu_evidence` (`src/tensor_grep/cli/agent_capsule.py:1549-1690`) | `_agent_gpu_tg_command()` -> `resolve_native_tg_binary()`, falls back to bare `"tg"` only if unresolved | `is_cross_domain_native_binary()` (`agent_capsule.py:1581`) | **Missed** the same shim case (same shared helper) |
| `tg calibrate` (`main.py:8197`) | `resolve_native_tg_binary()`, inherited stdio, no filesystem path argument passed | n/a | Not this bug class -- no Python-side path crosses the boundary |
| `tg devices` (`main.py:8204-8241`) | N/A -- pure Python `DeviceDetector`/`collect_device_inventory()`, no subprocess to a native binary at all | n/a | Not this bug class |

Both affected call sites share **one** classification helper
(`is_cross_domain_native_binary` / `native_binary_targets_windows`,
`src/tensor_grep/cli/runtime_paths.py:368-459` pre-fix), so the fix lands once and covers both.

## Root cause

`native_binary_targets_windows()` keys Windows-target detection on the `.exe` suffix of the
resolved candidate alone (`runtime_paths.py:368-378`, unchanged by this PR -- its narrow,
already-tested contract is preserved). That is correct for a raw binary, but both managed
installers also generate a **bare-named (`tg`, no extension) POSIX shim** for WSL/git-bash
shells that internally `exec`s the real `tg.exe`:

- `scripts/install.sh:363-377` -- the staged `bin/tg` heredoc, copied via `cp
  "$INSTALL_DIR/bin/tg" "$SHIM_PATH"` into every `SHIM_DIRS` entry (`install.sh:395-401`).
- `scripts/install.ps1:731-750` (`$frontdoorBashContent`) -- written once to the original
  `.tensor-grep/bin/tg`, then copied **byte-identically** into every `$shimDirs` entry
  (`~/.local/bin/tg`, `~/bin/tg`) via `Write-BashFile -Path $bashShimPath -Value
  $frontdoorBashContent` (`install.ps1:837-852`).

`resolve_native_tg_binary()` looks for a literal `tg` (no extension) on a WSL/Linux host
(`binary_name = "tg.exe" if sys.platform.startswith("win") else "tg"`, `runtime_paths.py:263`)
and legitimately returns **the shim**, not the `.exe` it delegates to. The `.exe`-suffix check
alone reads the shim as same-domain, so the probes hand it an untranslated `/tmp/...` path; the
shim's real Windows-target child process can't open it and returns the native binary's own
structured `path_not_found`, which the doctor then buckets as the honest-*looking* but wrong
`failed_probe_path` (same-domain bucket) instead of `failed_path_bridging`/`path_domain_mismatch`.

I initially assumed the sibling `tg-native-metadata.json` (already-shipped infra,
`native_frontdoor_metadata_path()`) would be sufficient to see through the shim -- it catches the
*original* `.tensor-grep/bin/tg`, but **not** the copies. `install.ps1:841`'s
`Copy-Item -LiteralPath $nativeFrontdoorPath -Destination $exeShimPath` copies a real `tg.exe`
into every `$shimDirs` entry alongside the shim, but does **not** copy `tg-native-metadata.json`
there -- and the shim-dir copies (`~/bin/tg`, `~/.local/bin/tg`), not the original front-door
directory, are what installer wiring puts on `$PATH`. Confirmed this is exactly what the
reporting box hit (see receipts below), so the fix adds a second, independent signal -- a
co-located `<name>.exe` on disk -- alongside the metadata check.

## Fix

Two new filesystem-observable signals in `runtime_paths.py`, OR'd into
`is_cross_domain_native_binary()` alongside the existing `.exe`-suffix check:

- `_native_frontdoor_metadata_targets_windows()` -- reads the sibling `tg-native-metadata.json`'s
  `asset_name` (present next to the *original* front door).
- `_sibling_native_windows_binary_exists()` -- a co-located `<name>.exe` file. **This is the one
  that actually catches the live repro** (the distributed shim-dir copies carry the `.exe` but not
  the metadata).

Both fail closed to `False` on any miss (no metadata file, no sibling `.exe`, unreadable,
malformed), so this can only ever **add** a cross-domain detection, never remove the pre-existing
one -- no risk to the `/mnt/<drive>/`-mounted same-domain Linux ELF case the original P0-1 fix
protects (covered by existing pinned tests, re-verified green below).

## WSL repro receipts (live, on the reporting box's real WSL environment)

**Before** (unmodified code) -- real `tg doctor --json`, `search_runtime_probe` section:
```json
{
  "status": "failed_probe_path",
  "command": "/mnt/c/Users/oimir/bin/tg search --gpu-device-ids 0 --json --no-ignore -F tg doctor gpu runtime probe <doctor-gpu-probe-file>",
  "exit_code": 2,
  "native_error_kind": "path_not_found",
  "error": "GPU runtime probe failed"
}
```
`resolve_native_tg_binary()` on this box resolves `/mnt/c/Users/oimir/bin/tg` -- a bash shim (no
`.exe`) that itself `exec`s `/mnt/c/Users/oimir/.tensor-grep/bin/tg` -> `tg.exe`.

**After** (this PR's `runtime_paths.py`, run against the same real resolved shim path via
`_doctor_gpu_search_runtime_probe`):
```json
{
  "status": "unsupported",
  "command": "/mnt/c/Users/oimir/bin/tg search --gpu-device-ids 0 --json --no-ignore -F tg doctor gpu runtime probe <doctor-gpu-probe-file>",
  "exit_code": 0,
  "routing_backend": "NativeCpuBackend",
  "routing_reason": "gpu-auto-fallback-cpu",
  "sidecar_used": false,
  "native_error_kind": null,
  "error": "GPU route did not use NativeGpuBackend (routing_backend=NativeCpuBackend, sidecar_used=False)."
}
```
The path now translates and opens cleanly through the real `tg.exe` (confirmed directly too: `tg.exe
search ... "$(wslpath -w probe.log)"` returns real match JSON, exit 0). The probe now reports the
*true* reason GPU acceleration is not available on this install -- no CUDA-enabled binary, matching
`tier.installed=false` -- instead of a spurious path failure. `_agent_gpu_evidence` (agent capsule),
which shares the same `is_cross_domain_native_binary()` gate, was verified fixed the same way.

## Validation

- `pytest tests/unit/test_runtime_paths.py tests/unit/test_agent_capsule_gpu_probe.py
  tests/unit/test_device_detect.py -q` -> 95 passed, 1 skipped (real-wslpath smoke test, skipped on
  native Windows/no-wslpath -- expected)
- `ruff check` / `ruff format --check --preview` / `mypy` on both touched files -> clean
- `git diff --check` -> clean
- All pre-existing pinned `is_cross_domain_native_binary`/`native_binary_targets_windows` tests
  (same-domain `/mnt/<drive>/` ELF, bare Linux CI runner, etc.) re-verified green, unmodified

One pre-existing, unrelated failure was observed and **ruled out** as caused by this change:
`test_cli_search_warns_when_gpu_device_id_out_of_local_inventory` (`test_cli_modes.py`) fails
identically against the pristine pre-PR `runtime_paths.py` (verified via a scoped `git stash` of
just that file) -- it is gated on `sys.platform.startswith("linux")`, which is `False` on the native
Windows test run, so this PR's code is provably never reached by that test. Out of scope (bootstrap
search-flow territory), left untouched.

## Scope note

GPU surface is experimental/default-OFF; this only affects the honesty of a diagnostic status
string on WSL cross-domain installs, not the fail-closed contract or any default-path behavior.
Did not touch the theoretically-similar but **unconfirmed** bare-`"tg"`-fallback gap in
`_agent_gpu_tg_command()` (only triggers when no native binary resolves at all) -- no live evidence
it is actually hit, and fixing it would need a different mechanism (pre-resolving via
`shutil.which` before classification); flagging here as a residual risk rather than speculatively
changing it.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/src/tensor_grep/cli/runtime_paths.py
+++ b/src/tensor_grep/cli/runtime_paths.py
@@ -378,16 +378,94 @@ def native_binary_targets_windows(binary: Path | str) -> bool:
     return str(binary).lower().endswith(".exe")
 
 
+# GPU-P0-1 follow-up (2026-07-21 CEO WSL dogfood, gotcontext-saddle): `native_binary_targets_windows()`
+# is keyed on the `.exe` suffix of the resolved candidate ITSELF, which is correct for a raw
+# binary but blind to the bare-named (`tg`, no extension) POSIX shim that BOTH managed installers
+# generate for WSL/git-bash/MSYS shells: `scripts/install.sh`'s staged `bin/tg` heredoc (copied via
+# `cp "$INSTALL_DIR/bin/tg" "$SHIM_PATH"` into every `SHIM_DIRS` entry, install.sh:395-401) and
+# `scripts/install.ps1`'s `$frontdoorBashContent` (`Write-BashFile -Path $stagingFrontdoorBashPath`
+# for the original `.tensor-grep/bin/tg`, then a byte-identical copy into every `$shimDirs` entry
+# via `Write-BashFile -Path $bashShimPath -Value $frontdoorBashContent`, install.ps1:837-852). Both
+# shims detect WSL via their own `/proc/version` "microsoft" check and then `exec` the real
+# `tg.exe` -- but the SHIM itself, not the `.exe` it delegates to, is what
+# `resolve_native_tg_binary()` returns on a WSL/Linux host (it looks for a literal `tg`, no
+# extension: `binary_name = "tg.exe" if sys.platform.startswith("win") else "tg"` above). The
+# `.exe`-suffix check alone therefore reads this shim as same-domain, so the doctor/agent probes
+# hand it an untranslated `/tmp/...` path; the shim's real Windows-target child process can't open
+# it and returns the native binary's own structured `path_not_found` -- which the doctor then
+# reports as the honest-LOOKING but WRONG `failed_probe_path` (same-domain bucket) instead of
+# `failed_path_bridging`/`path_domain_mismatch` (verified live: `tg doctor --json` under WSL
+# resolved a `~/bin/tg`-class shim (bash script, no `.exe`) that internally `exec`s `tg.exe`).
+#
+# Two independent signals below, both filesystem-observable from the shim's own resolved path:
+#   1. `_native_frontdoor_metadata_targets_windows()` -- the sibling `tg-native-metadata.json`
+#      written next to the ORIGINAL front door (`.tensor-grep/bin/tg-native-metadata.json`).
+#   2. `_sibling_native_windows_binary_exists()` -- a co-located `tg.exe`. This is the one that
+#      actually catches the live repro: install.ps1's `Copy-Item -LiteralPath $nativeFrontdoorPath
+#      -Destination $exeShimPath` (line 841) copies a REAL `tg.exe` into every `$shimDirs` entry
+#      alongside the bash shim, but does NOT copy `tg-native-metadata.json` there (only the
+#      original `.tensor-grep/bin/` gets it) -- so a shim resolved via `~/bin/tg` or
+#      `~/.local/bin/tg` (the common case; these, not `.tensor-grep/bin/`, are what installer
+#      wiring puts on `$PATH`) has signal 2 but NOT signal 1. Confirmed empirically: on the CEO's
+#      WSL dogfood box, `resolve_native_tg_binary()` resolved exactly such a shimDir copy, and
+#      re-running the doctor probe logic against it post-fix (signal 2 present, signal 1 absent)
+#      correctly flips `is_cross_domain_native_binary()` to True and the translated path opens
+#      cleanly via the real `tg.exe`. Keeping signal 1 too: it survives a `tg.exe` later being
+#      deleted out from under an otherwise-intact managed install (the shim's own fallback branch
+#      then execs the managed venv's Windows-side `python.exe`, still cross-domain).
+def _native_frontdoor_metadata_targets_windows(binary: Path | str) -> bool:
+    """True when the sibling `tg-native-metadata.json` records a Windows-target asset.
+
+    Reads `asset_name` (written by the installer next to whatever front-door entry point it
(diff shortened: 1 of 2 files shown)


oimiragieo:
Independent Opus gate: **SHIP-WITH-NITS** — NIT-2 (fail-closed + bounded metadata read) folded in the commit above. Documented-not-built per the gate's severity read: **NIT-1** — the sibling-`.exe` signal can misclassify a self-inflicted dual-build layout (in-tree Linux ELF + `tg.exe` in the same shared target dir on a WSL host); no installer produces this layout (install.sh names the Linux binary `tg-native`, never writes `tg.exe`), consequence is a GPU diagnostic flip only. **CRUX-4 follow-up** — `_agent_gpu_tg_command()`'s bare-`"tg"` fallback (agent_capsule.py:1507-1509) only fires when no native binary resolves at all; needs `shutil.which` pre-resolution in a separate small PR.

PR #706, opened 2026-07-22
fix(ledger): canonicalize claim/release/list to repo root, fix the PATH-scope footgun

oimiragieo:
## Summary

Fixes the #1 item from the CEO's v1.92.1 dogfood: `tg ledger claim core/hooks ...` then `tg ledger list` (or `list .`) returned an EMPTY claims list, and `tg ledger release` from a different PATH returned `released_count: 0` while the claim silently lived on. "Same-PATH-everywhere" was an undocumented sharp edge on the A2A coordination plane.

## The matching seam (cite file:line)

Pre-fix, `submit_claim`/`release_claim`/`list_claims` (`src/tensor_grep/cli/ledger_store.py:463,526,572` on `origin/main`) each independently did `root = _resolve_root(Path(path))` (imported from `session_store.py:345-347`) -- the CLI's literal `PATH` argument taken **at face value as the physical storage location** for `.tensor-grep/ledger/claims/index.json`. `claim core/hooks` and `list .` resolve to two different absolute directories (`<cwd>/core/hooks` vs `<cwd>`), so they wrote/read **two entirely different physical index.json files**. This is not a filtering bug -- `list_claims(".")` was reading a file that `claim core/hooks` never touched.

`_resolve_root` mirrors `session_store.py`/`checkpoint_store.py`'s own "root = wherever you point me" convention by design (verified: `checkpoint_store.py:196-233`'s `_detect_checkpoint_scope` deliberately does *not* walk a subdirectory up to the git top-level either -- checkpoint scope is legitimately variable). Ledger's coordination plane is different: its whole point is cross-invocation, cross-agent visibility for the *same* repository, so an unstated "you must invoke every command with the identical literal PATH string" contract is far more costly there.

## The fix

- **New `_discover_repo_root`/`_ledger_physical_root`** (`ledger_store.py`): canonicalize `root` to the nearest `.git` ancestor -- a pure filesystem `.exists()` walk, no subprocess -- falling back to today's literal-path behavior when no `.git` is found. A non-git working directory is not regressed; every existing test fixture (none of which `git init` by default) is byte-for-byte unaffected (verified -- all 88 pre-existing store tests + 30 pre-existing CLI tests + 4 concurrency tests pass unchanged).
- **`ClaimRecord` gains a `scope` field** -- the caller's original `PATH`, normalized root-relative POSIX (Windows separators, a leading `./`, a trailing `/` all fall out of `Path`+`.resolve()`+`.as_posix()` for free).
- **`list_claims` rolls scope UP**: listing a broader/ancestor path shows every claim scoped to it or to any descendant subtree (`_scope_contains`, segment-wise `PurePosixPath.parts` containment -- never a raw string-prefix test, so `core/ho` cannot false-match `core/hoodie`). Deliberately **one-directional**: a root-scoped claim does not roll DOWN into a narrower listing ("keep exact behavior for claims outside the listed subtree").
- **`release_claim` keeps its exact `--claim-id`/`--symbol` matching UNCHANGED** (I deliberately did *not* add rollup matching here -- it could let a release call silently drop an unrelated sibling agent's claim just because it shares a symbol name under a broader/ancestor scope). What changed: physical-root canonicalization means release now finds a claim filed from *any* subtree of the same repo (the "release right-path" dogfood case), and a genuine zero-match (wrong symbol/claim-id/agent) now returns `unmatched_reason` + a bounded (`_MAX_LIVE_CLAIMS_ELSEWHERE_SHOWN = 10`) `live_claims_elsewhere` list (`{claim_id, agent_id, scope, symbols, files, intent, expires_at}` each) instead of an indistinguishable `released_count: 0`.
- **Help text + docstrings** on `ledger_app`, `ledger_claim`, `ledger_release`, `ledger_list` state the canonicalization/rollup rule loudly (verified via a dedicated CLI test asserting `--help` output content).
- **Slice 2 (`record_finding`/`find_findings`) deliberately untouched** -- still plain `_resolve_root`, same footgun, not yet reported for that slice; called out explicitly in the module docstring, CONTRACTS.md, and `docs/multi_agent_context_plane.md` so it isn't mistaken for fixed.

## Dogfood repro, before -> after

```
$ tg ledger claim core/hooks --symbol open_session --agent-id agent-a
$ tg ledger list
BEFORE: {"claims": [], "count": 0}                         <- WRONG, claim invisible
AFTER:  {"claims": [{"scope": "core/hooks", ...}], "count": 1}

$ tg ledger release --symbol open_session --agent-id agent-a
BEFORE: {"released": [], "released_count": 0}               <- WRONG, claim lives on, no signal why
AFTER:  {"released": [{"scope": "core/hooks", ...}], "released_count": 1}

# genuine mismatch (wrong symbol) is now self-diagnosing instead of silent:
$ tg ledger release --symbol wrong-symbol --agent-id agent-a
AFTER:  {"released_count": 0, "unmatched_reason": "No live claim matched the given
         --claim-id/--symbol; 1 live claim(s) exist in this repository -- see
         live_claims_elsewhere.", "live_claims_elsewhere": [{"scope": "core/hooks", ...}]}
```
(Manually verified end-to-end against the real store functions in a throwaway git-inited temp repo before writing tests; see test list below for the automated equivalents, including a `monkeypatch.chdir` CLI-level test using the literal `claim core/hooks` / bare `list` argv the CEO described.)

## JSON additivity (contract-safety)

- `claim.scope` -- new key, additive.
- `release`'s 5 new keys (`listed_scope`, `unmatched_reason`, `live_claims_elsewhere`, `live_claims_elsewhere_count`, `live_claims_elsewhere_truncated`) -- additive; `None`/empty/`0`/`false` when `released_count > 0`, so a caller who only checks `released_count` sees no behavior change.
- `list_claims`'s **top-level** shape is unchanged (no new top-level key) -- I deliberately did *not* add a `listed_scope` top-level field to `list` (unlike `release`) specifically to avoid touching 5 currently-passing empty-result exact-dict-equality tests; only each claim's own dict content changed.
- `LEDGER_SCHEMA_VERSION` stays `1` -- this is an additive, non-breaking change (existing key semantics unchanged, only new keys added); a schema bump felt like overclaiming impact on Slice 2 (findings), which is untouched.
- Updated `docs/CONTRACTS.md` section 9 (claim shape, `release`/`list` response JSON, PATH-scoping paragraph) and `docs/multi_agent_context_plane.md` (line-number citations + a note that only Slice 1 is fixed). Verified: no governance test pins section 9's exact prose (`test_public_docs_governance.py`/`test_enterprise_docs_governance.py` grepped clean; both suites still pass, 58/58).

## Flags for the Opus gate

1. **Containment normalization correctness** -- `_scope_contains` uses `PurePosixPath(...).parts` slice comparison, not string prefix (guards the `core/ho` vs `core/hoodie` trap -- has a dedicated test). Windows-separator + `./`-prefix + trailing-`/` normalization has dedicated tests (`test_scope_normalizes_dot_slash_prefix_and_trailing_slash`, `test_scope_normalizes_windows_separators` -- the latter `skipif`'d to `win32` only, verified it runs and passes on this Windows dev box). `..`-escape is not a scope concern (that's `_normalize_relative_file`'s pre-existing `--files`-traversal guard, unchanged).
2. **Advisory exit-0 contract preserved** -- verified against `ledger_store.py`'s own docstrings + `docs/CONTRACTS.md` section 9 *before* building: `release_claim` never raises on a zero-match outcome; my `_release_result` helper only enriches the same non-raising return dict. `ledger_release`'s CLI wrapper's exit-code mapping (`main.py`) is untouched -- I only added JSON keys to the success payload.
3. **No store-format migration** -- `scope` is a brand-new field; an old on-disk record written by a pre-fix `tg` binary (missing the key entirely) is handled via `_entry_scope`'s backward-read default (`"."`, the maximally-visible choice) rather than a migration script -- has a dedicated test (`test_old_format_claim_missing_scope_defaults_to_root_and_stays_visible`) that hand-deletes the `scope` key from an on-disk record and confirms it still surfaces (not silently dropped).
4. **Physical-root canonicalization design choice** -- I considered (and rejected) reusing `repo_map._validation_repo_root` (an existing walk-up-to-project-root helper) because its npm/python/rust marker-file short-circuit would incorrectly stop the walk at a *nested* `package.json`/`pyproject.toml` inside a monorepo subtree (e.g. `packages/foo/package.json`), landing two different claims from two different monorepo packages on two different physical indexes -- exactly the bug this PR fixes, just relocated. `.git`-only detection (mirroring `checkpoint_store.py:196-233`'s established subprocess-based pattern, but as a pure filesystem walk instead, to keep `list_claims`/`release_claim` off the "shells out to git" cost path they didn't have before) avoids that trap.
5. **Ownership boundary respected** -- `tg prepare` / `main.py`'s prepare command was NOT touched (verified its `ledger_store.submit_claim` call site at `main.py:10386-10388` uses only pre-existing keyword args, and re-ran `tests/integration/test_prepare_oneshot_cuj.py` -- a pure-Python-subprocess test, no Rust build -- 9/9 pass). One pre-existing stale-after-my-edit line-number citation in a comment *inside* the prepare-owned code region (`main.py:10392`, citing `ledger_store.py:445-461`) was deliberately left untouched rather than risk a merge collision with the other agent's concurrent work on that file region -- flagging it here for a trivial follow-up fix (new location: the "NEVER blocks" docstring paragraph, now inside `submit_claim`'s docstring starting at line 592).
6. **A skill file** `.claude/skills/tensor-grep-ledger/SKILL.md` (documenting the now-fixed "PATH trap (1.92.1 dogfood)" as a known gotcha) exists only as an **uncommitted addition in the main checkout**, not in this worktree/branch -- outside this PR's diff entirely; flagging so a follow-up updates it once merged.

## Testing

- `ruff check` / `ruff format --check --preview` -- clean.
- `mypy --strict` on the **full** `src/tensor_grep` tree (81 files) -- clean.
- `pytest tests/unit/test_ledger_store.py tests/unit/test_ledger_cli.py tests/unit/test_ledger_concurrency.py` -- 129/129 pass (88 pre-existing + 41 new: rollup direction (both ways), disjoint-sibling isolation, false-prefix guard, Windows/`./`/trailing-slash normalization, release path-independence, release honesty fields, non-git sibling-isolation regression guard, backward-read for a pre-fix record, plus 7 new CLI-level tests including the literal `monkeypatch.chdir` + `claim core/hooks` + bare `list` repro).
- `pytest tests/integration/test_prepare_oneshot_cuj.py` -- 9/9 pass (verifies `tg prepare --claim`'s untouched integration point stays compatible).
- `pytest tests/unit/test_public_docs_governance.py tests/unit/test_enterprise_docs_governance.py` -- 58/58 pass (docs I edited aren't pinned elsewhere, and I didn't break anything that is).
- `pytest tests/unit/test_repo_map_targets.py tests/unit/test_agent_capsule_hardcases.py tests/unit/test_skill_index_sync.py` -- 49/49 pass (these reference "ledger_store"/"submit_claim" only as synthetic fixture strings, confirmed unaffected).
- Did **not** run `tests/e2e/test_routing_parity.py` (self-compiles Rust) or touch cargo/rustc/maturin, per the CPU-safe constraint.


oimiragieo:
Independent Opus gate: **SHIP-WITH-NITS** — all 4 cruxes verified empirically (21/22 e2e probes, worktree `.git`-FILE handling correct, 10/10 scope matrix incl. false-prefix + one-directionality, lock unified on the canonical root, advisory exit-0 + additive JSON preserved). **Nit #1 (migration honesty) folded** in the commit above — CONTRACTS §9 now states the pre-fix subtree-store orphaning + the ≤15-min release-mislead window + the safe manual cleanup. Nit #2 (stale citation main.py:10392, inside PR #705's owned region) and nit #3 (the uncommitted skill-file note in the main checkout) stay banked as post-merge follow-ups per the PR body.

PR #714, opened 2026-07-22
perf(prepare): reuse the capsule repo map for the blast-radius floor (truncation-gated)

oimiragieo:
opt10 campaign ranked-queue item #2 (`+10% overall improvement` campaign): stop `tg prepare` building the repo map twice.

## Verified seams (file:line, re-checked against this worktree before writing a line)

- `build_agent_capsule` (`src/tensor_grep/cli/agent_capsule.py:2675-2747`) builds `rm = repo_map.build_repo_map(path, max_repo_files=effective_max_repo_files, deadline_monotonic=deadline_monotonic)` (`:2729-2731`, `effective_max_repo_files` = `DEFAULT_AGENT_REPO_MAP_LIMIT` = 2000, `repo_map.py:156`), then delegates to `build_agent_capsule_from_map(rm, ..., _rescue_call_site_evidence=True)` (`:2732-2747`).
- `_build_prepare_payload` (`main.py:10438` in this branch, was `main.py:10375` pre-fix) called `build_agent_capsule(query, path, deadline_monotonic=deadline_monotonic)` — the outer wrapper, discarding `rm`.
- `_build_prepare_blast_radius_floor` (`main.py:10213` pre-fix) called `build_symbol_blast_radius(symbol, path, ...)` (repo_map.py:17248) whenever the query is natural-language (`call_site_evidence.status` is `skipped`/`error`, not `collected`/`collected_no_call_sites`/the specific "not found" reason) — and `build_symbol_blast_radius` itself calls `build_repo_map` again (`repo_map.py:17264-17269`) before delegating to `build_symbol_blast_radius_from_map` (`:17270-17277`). **Confirmed via direct call-count instrumentation**: for a natural-language prepare query on a non-truncated repo, `build_repo_map` ran exactly twice pre-fix.
- The map-reusing sibling `build_symbol_blast_radius_from_map(rm, symbol, ...)` (`repo_map.py:17438`) already exists and is the established `_from_map` pattern used elsewhere (`agent_capsule._collect_capsule_call_site_evidence_from_map`, `repo_map.py:821-935`; `main.py`'s own `blast-radius` command daemon gate, `main.py:11983-12010`).
- The truncation signal: `repo_map.build_repo_map` only sets `payload["scan_limit"]` when `max_repo_files is not None` (always true for this caller, capped at 2000), with `possibly_truncated` True only when the file-count cap actually dropped project files (`repo_map.py:6697-6716`).

All claims re-verified by reading the real code before implementing (not from memory/plan) — no discrepancies found vs. the ranked-queue item's own description.

## The fix

Two functions in `main.py`, **zero changes in `repo_map.py`/`agent_capsule.py`**:

1. **`_build_prepare_payload`**: builds `rm` directly (`build_repo_map(path, max_repo_files=DEFAULT_AGENT_REPO_MAP_LIMIT, deadline_monotonic=deadline_monotonic)`) — byte-identical to `build_agent_capsule`'s own body — then calls `build_agent_capsule_from_map(rm, query, deadline_monotonic=deadline_monotonic, _rescue_call_site_evidence=True)` instead of `build_agent_capsule(query, path, ...)`. The returned `capsule` is therefore byte-identical to what `build_agent_capsule` itself would have returned; `_build_prepare_payload` additionally keeps its own `rm` reference. **`build_agent_capsule`'s own signature/body/public return are completely untouched** — `tg agent` (`main.py:9599`) and the MCP `tg_agent` tool (`mcp_server.py:3089`) keep calling `build_agent_capsule` directly, unaffected.
2. **`_build_prepare_blast_radius_floor`**: takes a new required `rm` parameter. The former unconditional `build_symbol_blast_radius(symbol, path, ...)` call is now gated:
   - `rm['scan_limit']['possibly_truncated']` is **False** (the common case, any repo ≤2000 files) → reuse `rm` via `build_symbol_blast_radius_from_map(rm, symbol, max_depth=1, deadline_monotonic=deadline_monotonic)` + `_apply_blast_radius_output_limits(..., max_callers=8, max_files=8)` + `_copy_partial_signal(...)` — the exact same tail `build_symbol_blast_radius` applies internally, so output shape is unchanged.
   - `rm['scan_limit']['possibly_truncated']` is **True** (repo bigger than the 2000-file cap) → **preserve the exact pre-fix uncapped FS-backed rescan, unchanged, including its own comment explaining why `max_repo_files` must stay omitted.**

## THE LOAD-BEARING GUARD (why this isn't a blind reuse)

The floor's supplementary scan is deliberately uncapped (`max_repo_files` omitted, `main.py`'s own pre-existing comment) for large-repo recall, while the capsule's `rm` is capped at 2000 files. Reusing a *truncated* `rm` would silently narrow blast-radius recall on any repo bigger than the cap — a correctness regression, not just a missed speedup. The gate above makes that impossible: reuse only fires when `rm` is provably the complete file universe (nothing the uncapped scan could find that the capped one didn't already have), and falls through to the untouched fallback otherwise. This mirrors the exact TRAP-A pattern already documented and handled elsewhere in this codebase (`agent_capsule._collect_capsule_call_site_evidence_from_map`'s own docstring; `main._daemon_blast_radius_no_match_is_unreliable`) — same shared truncation signal, same falls-back-to-uncapped discipline.

## TDD (all 4 new tests confirmed RED against pre-fix `main.py` via `git stash`, then GREEN after)

New file: `tests/unit/test_prepare_blast_radius_floor_map_reuse.py`

1. `test_prepare_floor_reuses_map_when_not_truncated` — gate unit test (synthetic `rm`, monkeypatched `repo_map` functions): not-truncated routes to `build_symbol_blast_radius_from_map` and the FS-backed function is asserted **never called** (raises if it is).
2. `test_prepare_floor_preserves_fs_rescan_when_map_possibly_truncated` — the guard's negative half: possibly_truncated routes to the FS-backed function and `build_symbol_blast_radius_from_map` is asserted never called; also asserts the fallback never receives `max_repo_files`.
3. `test_prepare_builds_repo_map_once_and_matches_fs_backed_reference` — small real tree (the proven billing-shaped fixture from `tests/integration/test_prepare_oneshot_cuj.py`), full CliRunner `tg prepare` call with a `build_repo_map` call-counter: asserts **exactly 1** call (was 2), and that `blast_radius_floor.top_callers`' file set is byte-identical to calling the pre-fix FS-backed `build_symbol_blast_radius` directly against the same fixture/symbol — recall-identical, not just faster.
4. `test_prepare_recall_preserved_on_large_truncated_repo` — a **real, >2000-file synthetic tree** (2254 files: the billing shape + 50×45 padding dirs + the caller's file deliberately placed in a `zzz_caller/` directory that sorts after all padding, verified via `_iter_repo_files`'s documented bucket/round-robin walk order to land outside the 2000-file cap). Builds a genuinely-truncated `rm` via the real `build_repo_map`, then calls `_build_prepare_blast_radius_floor` directly (bypassing `tg prepare`'s own unrelated capsule-ranking/validation-detection machinery, which profiling showed varies ~16s-60s+ on this shared box independent of this change — see the test's own docstring). Asserts the caller living outside the cap window is still found (`callers_count >= 1`) and that exactly 2 `build_repo_map` calls occurred (the capped `rm` + the preserved uncapped fallback), proving the fallback genuinely ran.

Existing suites re-run green after the fix (all commands below via `PYTHONPATH=<worktree>/src C:/dev/projects/tensor-grep/.venv/Scripts/python.exe`, verified `tensor_grep.__file__` resolves into the worktree first, per the stale-venv trap):
- `tests/integration/test_prepare_oneshot_cuj.py` — 13/13 passed.
- `tests/unit/test_prepare_blast_radius_floor_map_reuse.py` (new) — 4/4 passed.
- `tests/unit/test_agent_capsule_*.py` + `test_agent_deadline_partial_stderr.py` + `test_agent_ignore.py` + `test_agent_reverse_import_graph_deadline_222.py` + `test_agent_suggested_scope.py` + `test_agent_vendored_subtree_{deadline_220,dedup_scale_222}.py` + `test_agent_workspace_root_advisory.py` — 199/199 passed (confirms `build_agent_capsule`'s own direct callers, `tg agent`, are unaffected).
- `tests/unit/test_cli_deadline_{coverage_gaps,flag,frontdoor_anchor}.py` + `test_refs_context_pack_deadline_205.py` + `test_repo_map_deadline.py` + `test_symbol_defs_stage_deadline_c1c2.py` + `test_blast_radius_mermaid.py` + `test_blast_radius_render_perf.py` + `test_context_render_and_blast_radius_max_files_bounds_suggested_edits.py` — 192/192 passed.

## Measure (A/B, frozen scorecard methodology)

Scorecard: `scratchpad/opt10/scorecard_definition.md`; corpus A = `scratchpad/opt10/corpora/tg-pinned` (pinned v1.93.2 clone, HEAD `4e471cc`, confirmed unmodified in `src/tensor_grep/cli/{main,repo_map,agent_capsule}.py` before use). S5 cell: `prepare <A>/src "add a flag to tg prepare" --json`. Baseline (published wheel, `uvx`) = **10.574s median**.

This PR's A/B is SOURCE-tree before/after, same interpreter (`C:/dev/projects/tensor-grep/.venv/Scripts/python.exe`), `PYTHONPATH`-swapped between the unmodified pinned clone (before) and this worktree (after), invoked as `python -m tensor_grep prepare <A>/src "add a flag to tg prepare" --json` with `TG_SESSION_DAEMON_AUTOSTART=0`. 5 cold reps/arm, serial, `tensor_grep.__file__` verified to resolve into the correct source before timing anything.

| | median_s | min_s | max_s | spread_s |
|---|---|---|---|---|
| before (pinned v1.93.2, unmodified) | 11.4455 | 11.2911 | 11.7024 | 0.4113 |
| after (this branch) | 11.0196 | 10.3981 | 11.7187 | 1.3206 |

Delta: 0.4259s / **3.72%** — this does **not** clearly exceed the combined noise spread (1.3206s), so per this campaign's own noise-floor discipline I am reporting it as **noise-level at the wall-clock/subprocess level on this shared box, not a confirmed end-to-end speedup** — not overclaiming a win the numbers don't clearly support.

Separately, the **mechanism is unambiguously confirmed** via direct `build_repo_map` call-count instrumentation on this exact corpus/query (in-process, no subprocess variance): **2 calls before this fix, 1 call after** (isolated single-shot timings: ~21.2s before vs ~7.2s after total wall for `_build_prepare_payload`, consistent in direction with the median result but far noisier at N=1, not cited as the headline number). Corpus A's `/src` target is modest in scope (`scan_limit.scanned_files=81` on this query), so the absolute avoided-rescan cost here (~2-4.5s of `build_repo_map` time observed directly) is real but small relative to this specific shared box's run-to-run jitter; the ranked-queue item's own predicted ~28% figure was derived under different profiling conditions and should be read as directional, not confirmed by this specific A/B.

## Scope discipline

- Changes confined to `main.py`'s two functions named in the task (`_build_prepare_blast_radius_floor`, `_build_prepare_payload`). Zero changes to `repo_map.py` or `agent_capsule.py`.
- `build_agent_capsule`'s public return is untouched — `tg agent` / MCP / their existing tests are unaffected (confirmed via the 199-test agent-capsule suite run above).
- `ruff check .`, `ruff format --check --preview .` (both clean on the 2 touched files; 4 pre-existing unrelated `.md`-fence drift files elsewhere in the repo, not touched by this PR), `mypy src/tensor_grep` (81 files, no issues) all green.

## Flag for independent Opus gate

Recall-preservation is the crux of this change: a blind reuse of a truncated map would be a silent blast-radius recall regression on any repo over 2000 files, and would not be caught by any existing test (none of the pre-existing suites exercise the >2000-file/`possibly_truncated=True` path through `tg prepare`). Please verify:
1. The `possibly_truncated` gate in `_build_prepare_blast_radius_floor` (`main.py`) is evaluated **before** any reuse attempt, on every code path that reaches it.
2. The fallback branch is byte-identical to the pre-fix code (diff it against `origin/main`'s copy).
3. `build_agent_capsule`'s signature/body genuinely has zero diff (nothing in `agent_capsule.py` changed).
4. Test 4's fixture placement claim (the caller genuinely lands outside the 2000-file scan window) — re-run it and inspect `rm['scan_limit']` yourself rather than trusting the docstring's math.

### DIFF
--- a/src/tensor_grep/cli/main.py
+++ b/src/tensor_grep/cli/main.py
@@ -10213,6 +10213,7 @@ def route_test(
 def _build_prepare_blast_radius_floor(
     *,
     path: str,
+    rm: dict[str, Any],
     target: dict[str, Any],
     call_site_evidence: dict[str, Any],
     related_call_sites: list[dict[str, Any]],
@@ -10225,11 +10226,25 @@ def _build_prepare_blast_radius_floor(
     (agent_capsule.py:536,546) -- a natural-language task query fails that gate and leaves
     ``related_call_sites`` EMPTY even though the SELECTED primary target may have real callers.
     Reuse the capsule's own scan when it already ran (``status`` ``collected``/
-    ``collected_no_call_sites``); otherwise run exactly ONE supplementary FS-backed scan via the
-    literal-seed-rescue-equipped ``build_symbol_blast_radius`` (never the rescue-less
-    ``_from_map`` sibling -- see ``_collect_capsule_call_site_evidence_from_map``'s own TRAP A
-    docstring: a no_match against a possibly-truncated map with no rescue reads as a false-
-    complete absence) keyed on the symbol the capsule actually selected, not the raw query text.
+    ``collected_no_call_sites``); otherwise fall to a supplementary blast-radius scan keyed on
+    the symbol the capsule actually selected, not the raw query text.
+
+    opt10 campaign ranked-queue item #2: that supplementary scan used to ALWAYS pay a second
+    FS-backed ``build_repo_map`` walk+parse (via ``build_symbol_blast_radius``) even though
+    ``_build_prepare_payload`` already built one moments earlier and now passes it in as ``rm``
+    -- doubling prepare's dominant cost on every natural-language query (the common CUJ). Reuse
+    ``rm`` via the map-reusing sibling ``build_symbol_blast_radius_from_map`` UNLESS ``rm``'s own
+    truncation signal (``rm['scan_limit']['possibly_truncated']``, repo_map.py:6707-6715) says
+    the capsule's ``DEFAULT_AGENT_REPO_MAP_LIMIT``-capped map may be missing files a full scan
+    would find. THE LOAD-BEARING GUARD: the supplementary scan below is DELIBERATELY uncapped
+    (``max_repo_files`` omitted, see that branch's own comment) for large-repo recall, while
+    ``rm`` is capped at 2000 files -- blindly reusing a truncated ``rm`` would silently narrow
+    blast-radius recall on a >2000-file repo, a real correctness regression, not just a missed
+    speedup. When ``rm`` is complete (any repo at or under the cap, the common case), it already
+    IS the same file universe an uncapped rescan would produce, so reuse is recall-identical, not
+    just faster -- never stamps the rescue-less daemon ``TRAP A`` sentinel (see
+    ``_collect_capsule_call_site_evidence_from_map``'s docstring) because a no_match on a
+    ``possibly_truncated`` map is impossible on this branch by construction of the gate above it.
 
     Returns ``(floor, deadline_partial)``. ``deadline_partial`` is deliberately narrower than the
     floor's own ``possibly_incomplete`` field: it is True ONLY when a ``--deadline`` cutoff
@@ -10239,7 +10254,12 @@ def _build_prepare_blast_radius_floor(
     (``_scan_incomplete``'s own documented contract).
     """
     from tensor_grep.cli.agent_capsule import _related_call_site_record
-    from tensor_grep.cli.repo_map import build_symbol_blast_radius
+    from tensor_grep.cli.repo_map import (
+        _apply_blast_radius_output_limits,
+        _copy_partial_signal,
+        build_symbol_blast_radius,
+        build_symbol_blast_radius_from_map,
+    )
 
     symbol = str(target.get("symbol") or "")
     status = str(call_site_evidence.get("status") or "")
@@ -10275,6 +10295,38 @@ def _build_prepare_blast_radius_floor(
         if error is not None:
             floor["error"] = error
         return floor, bool(deadline_partial)
+
+    def _floor_from_radius_payload(radius_payload: dict[str, Any]) -> tuple[dict[str, Any], bool]:
+        # Shared tail for BOTH the from-map reuse branch and the FS-backed fallback branch below
+        # -- one definition so the two paths cannot silently drift apart on how a radius_payload
+        # becomes a floor.
+        if radius_payload.get("no_match"):
+            return _floor(
+                source="supplementary_blast_radius",
+                related=[],
+                graph_trust_summary=dict(radius_payload.get("graph_trust_summary") or {}),
+                resolution_gaps=list(radius_payload.get("resolution_gaps") or []),
+                deadline_partial=bool(radius_payload.get("partial")),
+                omitted=0,
+            )
+        related = [
+            record
+            for record in (
+                _related_call_site_record(caller, target_symbol=symbol)
+                for caller in (radius_payload.get("callers") or [])
+                if isinstance(caller, dict)
+            )
+            if record is not None
+        ]
+        output_limit = radius_payload.get("output_limit") or {}
+        return _floor(
+            source="supplementary_blast_radius",
+            related=related,
+            graph_trust_summary=dict(radius_payload.get("graph_trust_summary") or {}),
+            resolution_gaps=list(radius_payload.get("resolution_gaps") or []),
+            deadline_partial=bool(radius_payload.get("partial")),
+            omitted=int(output_limit.get("omitted_callers") or 0),
+        )
 
     if status in ("collected", "collected_no_call_sites"):
         return _floor(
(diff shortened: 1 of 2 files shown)


oimiragieo:
**Gate NIT-1 resolved — test-fidelity fix pushed (0c3ec46).**

The independent Opus gate verdict was **SHIP-WITH-NITS**: the routing (test 2 gates it directly) and the load-bearing truncation guard are sound; the one nit was that test 4 (`test_prepare_recall_preserved_on_large_truncated_repo`) did not actually place its caller *outside* the 2000-file capped scan window — the repo-map walk is round-robin by directory, so a lone file in a 1-file `zzz_caller/` dir lands in round 0 and **survives** the cap. The caller was in-window, a naive always-reuse would have found it too, and the recall claim went untested.

Fixed:
- **Bury the caller out-of-window deterministically** — `zzzz_run.py` now sits at within-dir sorted index ≥ 2000 behind `_LARGE_TREE_BURY_FILE_COUNT` (2100) `pad#####.py` files. The walk draws ≤ 1 file/dir/round, so max round before the 2000-file cap is ≤ 2000; any file at within-dir index ≥ 2000 is **guaranteed** dropped, independent of sibling-dir count. The definition (`billing.py`, top-level group-2) stays in-window so the symbol still resolves.
- **Added the gate-requested contrast assertion** — a naive from-map reuse (`build_symbol_blast_radius_from_map` on the truncated `rm`, `max_depth=1`, exactly as the floor's reuse branch calls it) **MISSES** the buried caller, while the guarded floor still **FINDS** it via the preserved uncapped rescan. Both asserted in the same run → the test is now non-vacuous.
- Dropped the orphaned `_LARGE_TREE_PADDING_DIR_COUNT`/`_LARGE_TREE_PADDING_FILES_PER_DIR` constants; rewrote the fixture docstring to match.

Local verify (pre-built `.venv` + `PYTHONPATH=<worktree>/src`, `tensor_grep.__file__` confirmed resolving into the worktree): **4/4 pass (31.5s)**; `ruff check` + `ruff format --check --preview` clean; CI mypy scope is `src/tensor_grep` only (tests excluded), so the pre-existing test-signature `no-untyped-def` notes are out of gate.

Marking ready for review.

Issue #48, opened 2026-05-07
perf: reduce public shim startup overhead against rg and ast-grep baselines


oimiragieo:
Follow-up blocker for a future world-class/100% ready claim.

Context:
- Public CLI shim remains materially slower than raw `rg` for cold exact-text search and slower than `ast-grep` for public-shim AST startup.
- Keep current positioning honest: `rg` is the cold exact-text baseline, `ast-grep` is the structural-search baseline, and `tg` is the agent-native orchestration layer.

Acceptance direction:
- Add benchmark evidence for public installed CLI startup/control-plane changes.
- Do not update speed claims without accepted benchmark artifacts.
- Preserve semantic parity while reducing startup overhead.

oimiragieo:
Update after v1.13.7: PR #212 shipped the benchmark/provenance prerequisite for this issue: launcher command kind is recorded per cold-path row, shim/Python launcher warnings are aggregated, and stale in-tree native binaries are blocked by default unless explicitly exploratory. That makes future startup/control-plane work measurable without mixing native and wrapper timings. Leaving this issue open because the actual public shim/startup overhead reduction has not yet been proven by accepted benchmark artifacts.

oimiragieo:
Update after v1.13.11: the v1.13.8 dogfood follow-ups landed in PRs #217-#220 and are published through v1.13.11, but this startup/control-plane issue remains open. Current repo evidence in docs/world_class_plan.md still says the prior launcher/control-plane probes (explicit_binary_early_rg, explicit_binary_positional_early_rg, and tg-search-fast) were rejected by benchmark artifacts; the next honest fix here is a larger native control-plane rewrite with accepted benchmark proof, not another claim-only launcher toggle.

oimiragieo:
## Benchmark evidence (v1.69.3, Windows, best-of-6 on a real repo, flask/src/flask)

| command | median |
|---|---|
| `tg --version` (pure shim startup) | 193 ms |
| `tg search <literal>` (shim + search) | 232 ms |
| `rg <literal>` (raw baseline) | 22 ms |
| `ast-grep run` | 35 ms |

The public shim carries **~180 ms** of startup overhead (~9x `rg`, ~5x `ast-grep`). Prior import-defer work reduced this from ~270 ms, but a material gap remains.

**`python -X importtime` profile (self-time, sums to ~174 ms):** the cost is **diffuse -- no single dominant import to defer**:
[...]

PR #715, opened 2026-07-22
perf(cold-start): defer heavy imports off the fast paths (single-sourced scan limits)

oimiragieo:
## Summary

Ranked-queue item **#6** of the +10% campaign (source: `docs/planning` optimization queue) — the
import-deferral cold-floor bundle, three independently-scoped moves (`F2.6` + `F2.5` + `F2.4`).
**Results-identical perf**; `F2.4` is contract-touching (single-source-of-truth constants), flagged
below for a light independent check.

### F2.6 — `cli/__init__.py:2` (`from typing import Any`)

The sole use of `Any` was `__getattr__`'s return annotation (`cli/__init__.py` is the module-level
`__getattr__` (PEP 562) hook exposing `tensor_grep.cli.main_entry` without eagerly importing
`bootstrap`). Fixed via `-> object` instead (needs no import) — **not** `from __future__ import
annotations` alone, which I initially tried and is wrong here: PEP 563 only defers *evaluation* of
an annotation to a string; a static checker still has to *resolve* that string against the module's
real bindings. Confirmed empirically —

```
$ mypy cli/__init__.py   # -> Any, `from __future__ import annotations`, no `typing` import
error: Name "Any" is not defined  [name-defined]
```

`object` needs no import at all and is a pure static-typing change (`__getattr__`'s PEP 562
dispatch is untyped at runtime either way). Verified nothing in-repo relies on the wider `Any`
static type: the real console-script entry point in `pyproject.toml` is `tensor_grep.cli.bootstrap:
main_entry` directly, bypassing this `__getattr__` hook entirely.

`cli/__init__.py` is the parent-package init for every `tensor_grep.cli.*` submodule, so its cost
is paid on **every** `tg` invocation before any submodule-specific code runs.

### F2.5 — `runtime_paths.py:1,4` (`import json` / `import shutil`)

Both moved to function-local at their only call sites: `json` → `_read_native_frontdoor_metadata`;
`shutil` → `translate_path_for_windows_binary` and `resolve_ripgrep_binary` (two separate call
sites, two separate local imports, matching the file's existing style). `cli/bootstrap.py` imports
this whole module eagerly (`from tensor_grep.cli.runtime_paths import (env_flag_enabled,
resolve_native_tg_binary, resolve_ripgrep_binary)`) for **every** invocation including `--version`,
which calls none of the three json/shutil-using functions (confirmed: `bootstrap._print_version`
has its own separate `_read_project_version_fallback`, never touches `runtime_paths`' version of
it).

### F2.4 (contract-touching) — `main.py:46-52`'s 5 `directory_scanner` constants

New zero-dependency module `src/tensor_grep/io/scan_limits.py` holds the 5 constants
(`UNBOUNDED_VENDORED_ROOT_DIR_NAMES`, `BROAD_WORKSPACE_PROJECT_MARKERS`,
`BROAD_WORKSPACE_PROJECT_CHILD_THRESHOLD`, `BROAD_WORKSPACE_MARKED_ROOT_CHILD_THRESHOLD`,
`IMPLICIT_SEARCH_WALK_FILE_CEILING`) plus the private `_GENERATED_DIR_NAMES` skip-list
`UNBOUNDED_VENDORED_ROOT_DIR_NAMES` is derived from (moved together so the derivation and its
input stay in one place — no second copy to drift). **Zero imports** — just frozenset/int literals.

- `io/directory_scanner.py` re-exports every name (`from tensor_grep.io.scan_limits import X as X`
  — the same redundant-alias idiom `cli/bootstrap.py` already uses for `run_subprocess`), so its
  public surface is byte-identical for any existing `from tensor_grep.io.directory_scanner import
  <NAME>` caller.
- `cli/main.py:46-52`'s module-level import now points at `scan_limits` directly.
- `cli/bootstrap.py`'s 3 guard helpers (`_path_has_project_marker`,
  `_search_paths_include_workspace_root`, `_search_paths_include_vendored_root`) now do
  `from tensor_grep.io.scan_limits import ...` instead of `...directory_scanner import ...`. The
  **4th** helper, `_search_paths_include_oversized_implicit_root`, is deliberately **unchanged** —
  it constructs a real `DirectoryScanner`, so it genuinely needs the heavier module.
- `cli/scan_guardrails.py` (the `tg scan`/`test`/`ast-info`/`run` guard, reached via the lazily
  imported `cli/ast_workflows.py`) repointed the same way — it only ever needed the 3 plain
  constants, never `SearchConfig`/`DirectoryScanner`.
- The Rust-parity comment chain (`IMPLICIT_SEARCH_WALK_FILE_CEILING`'s cross-reference to
  `rust_core/src/rg_passthrough.rs`'s own `pub const IMPLICIT_SEARCH_WALK_FILE_CEILING: usize =
  1500`) moved intact into `scan_limits.py` — Rust untouched, verified the numeric value still
  matches by reading `rg_passthrough.rs:153` directly (no cargo/rustc/maturin run).

**Why the win is bigger than "5 constants": before this fix, `bootstrap.py`'s broad-scan guard
*always* imported the whole `directory_scanner` module (hence `SearchConfig`/`dataclasses`/
`inspect`) just to read 3 frozen/int constants — even when the search was ultimately delegated to
the native binary or `rg` (both walk in a **separate process** and never touch Python's
`DirectoryScanner` at all). That's the `explicit-path search` cell below: a full, clean elimination
of `directory_scanner`+`core.config`, not just a smaller module swap.**

## Verify-first findings (two claims narrowed after checking the real import graph)

1. **`cli/formatters/json_fmt.py:5`** has its own **unrelated, pre-existing** module-level
   `from tensor_grep.core.config import SearchConfig` (reached via `cli/main.py`'s own eager
   `from tensor_grep.cli.formatters.base import OutputFormatter`, which pulls in the whole
   `formatters` package). Traced via a `builtins.__import__` hook, not guessed. Consequence: a bare
   `cli.main` import (simulating `--help`) still loads `core.config` — just no longer *via*
   `directory_scanner`. This PR does not touch it (out of scope: F2.4 is about the 5 broad-scan
   constants' source module, not every `SearchConfig` import in the CLI); noted here instead of
   silently folding it into the headline claim.
2. **`cli/subprocess_policy.py:7`** has an *identical* redundant `from typing import Any` — also
   eagerly loaded by `cli/bootstrap.py` — which fully masks F2.6's saving for the whole-process
   `--version` case (typing loads via this path regardless of the `cli/__init__.py` fix). I
   attempted the same fix there (`**kwargs: object`, drop the import) and **reverted it**:
   `subprocess.run`'s typeshed stub is overload-heavy and does not accept an `object`-typed
   `**kwargs` splat — `mypy --strict` produced 2 real errors (`no-any-return`, `call-overload`), not
   a safe drop-in. Left untouched; noted rather than silently shipping a smaller win than claimed.

## TDD

- **(a)** Extended `tests/unit/test_bootstrap_fast_path_imports.py` (the existing house pattern:
  subprocess-based `sys.modules` probes, deliberately not in-process) with **11 new tests** — one
  absence pin per deferred-import claim (`scan_limits` present/`directory_scanner`+`core.config`
  absent for the 3 guard helpers called directly, for a bare `cli.main` import, for a bare
  `cli.scan_guardrails` import, and for `scan_limits` itself in isolation; `json`/`shutil` absent
  for a bare `runtime_paths` import), plus 2 non-regression behavioral checks (`--version`/`-V`
  still print the banner and exit 0; the `__getattr__` hook still resolves `main_entry` correctly
  after the `Any`→`object` change) and a source-level pin (`cli/__init__.py` has no eager `import
  typing`). One existing test (`test_plain_search_still_imports_directory_scanner_when_actually_
  needed`) needed updating: pre-fix, `directory_scanner` loaded unconditionally for any
  search-shaped argv; post-fix it loads only when Python's own walk is genuinely needed, so the
  test now explicitly forces the "no native, no rg" branch (mirroring `test_cli_bootstrap.py`'s own
  in-process monkeypatch pattern, via subprocess source-injection) instead of depending on the
  ambient dev machine's PATH.
- **(b)** Added `test_scan_limits_single_source_of_truth_across_every_re_export` to
  `test_cli_bootstrap.py`: asserts `scan_limits`'s values `is` (not just `==`) identical to
  `directory_scanner`'s re-export, `cli/main.py`'s private aliases, and `cli/scan_guardrails.py`'s
  private aliases — `is` makes the single-source-of-truth contract a **build-time fact** (an
  accidental future re-literal instead of a re-import would fail this test even though `==` would
  still pass on equal-but-distinct frozensets). Re-pointed the ~13 existing
  `from tensor_grep.io.directory_scanner import <CONSTANT>` imports in `test_cli_bootstrap.py` to
  `scan_limits` (grepped every constant name across `tests/`, per the task).
- **(c)** Full behavior: ~950 tests across the bootstrap / `directory_scanner` /
  `scan_guardrails`-adjacent (`ast_workflows`, `r2_dos_batch`, `cli_modes`) / orient (7 files) /
  ast / routing / `medlow_main_cli` suites, all green **except 3 pre-existing failures** — verified
  byte-identical against this worktree's own unmodified `HEAD` via `git stash` (a missing compiled
  `rust_core` PyO3 extension in this fresh worktree; unrelated to this Python-only change; never
  ran cargo/rustc/maturin per the CPU-SAFE constraint). `tests/e2e/test_routing_parity.py` was never
  run (it self-compiles Rust).

## Measured: `-X importtime` A/B (`python -m tensor_grep`, PYTHONPATH swap)

Before = the pinned `HEAD` (v1.93.2 lineage) source reconstructed into a scratch tree via
`git show HEAD:<path>` for each of the 6 touched files; after = this worktree. **6 reps/side,
interleaved rep-by-rep** (before-rep-N, after-rep-N, ...) rather than block-sequential — this
machine measured **~74% background CPU load** from concurrent campaign agents during the run, and
interleaving keeps a transient load spike from skewing one arm. One discarded warm-up rep/side
first equalizes OS disk-cache state (a freshly-materialized source tree pays a one-time cold-cache
tax on its first launch that has nothing to do with the code — an earlier block-sequential run
showed a 633ms outlier on `before`'s first `--help` rep purely from this).

**Median total self-time** (sum of every module's own self-time in the trace — noisy at the
whole-process level on this shared/busy machine; the per-module cumulative deltas below are the
trustworthy signal):

| command | before (median, n=6) | after (median, n=6) | delta |
|---|---|---|---|
| `--version` | 164.45ms | 154.51ms | **-9.93ms** |
| `--help` | 210.75ms | 199.36ms | **-11.40ms** |
| `explicit-path search` | 165.08ms | 146.33ms | **-18.75ms** |

**Per-module cumulative time** (median, ms) — the mechanistic evidence, and confirmation the
deferred modules are absent from the after-trace where claimed:

| module | `--version` before→after | `--help` before→after | `explicit-path search` before→after |
|---|---|---|---|
| `tensor_grep.io.directory_scanner` | absent both (prior #48 fix) | **1.875 → absent** | **13.568 → absent** |
| `tensor_grep.core.config` | absent both | 6.370 → 7.316 (unchanged — the `json_fmt.py` residual above) | **10.819 → absent** |
| `tensor_grep.io.scan_limits` | n/a (not on this path) | absent → **1.159** (new, cheap) | absent → **1.076** (new, cheap) |
| `tensor_grep.cli.runtime_paths` | **15.332 → 0.587** | **13.671 → 0.590** | **13.826 → 0.631** |
| `tensor_grep.cli.bootstrap` | 60.343 → 42.643 | 57.790 → 40.597 | 58.227 → 42.562 |
| `tensor_grep.cli.main` | n/a | 124.506 → 125.707 (unchanged — `core.config` still loads via `json_fmt.py`, netting the small `directory_scanner`/`scan_limits` swap to ~noise) | n/a |

**Honest read**: `runtime_paths` (F2.5) is a consistent, unconditional win on **every** command —
~13-15ms cumulative recovered regardless of what the command does, because `bootstrap.py` always
imports it. F2.4's win is command-shape-dependent: full and clean on `explicit-path search` (both
`directory_scanner` and `core.config` fully eliminated, ~18.75ms total — the case the campaign item
was written for), partial on `--help` (`directory_scanner` eliminated but `core.config` persists
via the unrelated `json_fmt.py` path, so `cli.main`'s own cumulative is a wash). F2.6 has no
directly-attributable line in this trace (`typing` still loads via `subprocess_policy.py`, see
above) — its win is the qualitative one (one fewer dead-weight import statement on the hottest
path in the codebase), not a quantifiable ms figure from this measurement. Subtracting nothing:
these are the raw numbers, on a busy shared machine, with the caveats stated.

## Flag for a light independent check

**F2.4's shared-constant parity** — please re-verify independently:
1. `scan_limits.py`'s 5 constants have the exact same values as `directory_scanner.py` had before
   this PR (diff the two module bodies at the relevant lines).
2. `IMPLICIT_SEARCH_WALK_FILE_CEILING = 1500` still numerically matches `rust_core/src/
   rg_passthrough.rs:153`'s `pub const IMPLICIT_SEARCH_WALK_FILE_CEILING: usize = 1500` (Rust file
   untouched by this PR — comment-verified only, no build run).
3. The sync-pin test (`test_scan_limits_single_source_of_truth_across_every_re_export`) actually
   fails if you revert `directory_scanner.py`'s re-export to a hand-copied literal instead of an
   import (a cheap way to confirm the `is`-based test has teeth, not just `==`).

## Gates

- `ruff check .` (repo-wide): clean.
- `ruff format --check --preview .` (repo-wide, the real CI scope): clean on every file this PR
  touches; 4 **pre-existing** doc-fence diffs elsewhere (`docs/architecture.md`,
  `docs/planning/PROJECT_PLAN.md`, 2 files under `docs/plans/`) — confirmed present on `HEAD`
  unmodified too (known, separate whole-repo markdown-fence gap, not introduced by this PR).
- `mypy src/tensor_grep` (the real CI scope, 82 files): clean.
- New/updated tests: 13 in `test_bootstrap_fast_path_imports.py`, +1 sync-pin in
  `test_cli_bootstrap.py` (94 tests total in that file, all passing).

CPU-SAFE: no cargo/rustc/maturin run at any point; Rust files read-only. `tests/e2e/
test_routing_parity.py` never run. `main.py`'s only edit is the `:46-52` import block (verified via
`git diff --stat` before committing — no overlap with the concurrent PREPARE-region agent working
around `:10200-10600`).

### DIFF
--- a/src/tensor_grep/cli/__init__.py
+++ b/src/tensor_grep/cli/__init__.py
@@ -1,10 +1,25 @@
+# perf (+10% campaign #6 / F2.6): `tensor_grep.cli` is the PARENT PACKAGE of every `tensor_grep.
+# cli.*` submodule (`bootstrap`, `main`, `runtime_paths`, ...), so THIS file's own top-level import
+# cost is paid on LITERALLY EVERY `tg` invocation -- including the trivial `--version` fast path --
+# before any submodule-specific code ever runs. `from typing import Any` used to be the ONLY
+# reason `typing` (a measurably non-free stdlib import) loaded this early; the sole use below,
+# `__getattr__`'s return annotation, never needs to be `Any` specifically -- returning the builtin
+# `object` instead needs no import at all (verified: nothing in this repo does `from tensor_grep.
+# cli import main_entry` and calls it -- the real console-script entry point in pyproject.toml
+# targets `tensor_grep.cli.bootstrap:main_entry` directly, bypassing this `__getattr__` hook
+# entirely -- so narrowing this hook's static return type has no in-repo call-site impact).
+# NOTE: `from __future__ import annotations` alone would NOT have been sufficient here -- PEP 563
+# only defers *evaluation* of an annotation to a string; mypy still resolves that string against
+# the module's real (possibly TYPE_CHECKING-gated) imports, and a bare `from typing import
+# TYPE_CHECKING` pays the exact same `typing`-module-load cost this fix exists to avoid. Verified:
+# `mypy --strict` on this file with `-> Any` and no `typing` import at all fails with `Name "Any"
+# is not defined`; `-> object` needs no import and passes clean.
 from importlib import import_module
-from typing import Any
 
 __all__ = ["main_entry"]
 
 
-def __getattr__(name: str) -> Any:
+def __getattr__(name: str) -> object:
     if name != "main_entry":
         raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
     value = import_module("tensor_grep.cli.bootstrap").main_entry
--- a/src/tensor_grep/cli/bootstrap.py
+++ b/src/tensor_grep/cli/bootstrap.py
@@ -27,9 +27,21 @@ from tensor_grep.cli.subprocess_policy import run_subprocess as run_subprocess
 # transitively pulls in the stdlib `dataclasses` module -- and `dataclasses` imports `inspect`
 # (plus `ast`/`dis`/`tokenize`/`copy`/`weakref`), a chain measured at ~25ms cumulative
 # (`python -X importtime -c "import tensor_grep.cli.bootstrap"`, warm cache) that a trivial
-# `--version` call has no reason to pay. Each of the 3 helper functions does its own
-# function-local `from tensor_grep.io.directory_scanner import ...` instead; see
-# `tests/unit/test_bootstrap_fast_path_imports.py` for the regression test that pins this.
+# `--version` call has no reason to pay.
+#
+# perf (+10% campaign #6 / F2.4, follow-up): each of the 3 helper functions used to do its own
+# function-local `from tensor_grep.io.directory_scanner import ...` -- which, being function-local,
+# already kept the cost OFF the `--version` fast path (this docstring's original claim), but still
+# paid the full `SearchConfig`/`dataclasses`/`inspect` chain the MOMENT any of the three actually
+# ran (i.e. on every real search invocation that reaches the broad-scan guard, REGARDLESS of
+# whether the search ultimately gets delegated to the native binary or `rg` -- both of which do
+# their own walk in a separate process and never touch Python's `DirectoryScanner` at all). Each
+# helper now does `from tensor_grep.io.scan_limits import ...` instead -- the same zero-dependency
+# module `cli/main.py`'s module-level broad-scan literals read from -- so the guard-only path no
+# longer drags in `SearchConfig` at all; `directory_scanner` is still imported exactly when a real
+# Python-side walk needs it (e.g. `_search_paths_include_oversized_implicit_root` below, which
+# constructs a real `DirectoryScanner`). See `tests/unit/test_bootstrap_fast_path_imports.py` for
+# the regression tests that pin both claims.
 
 # Saved at import time so _streaming_passthrough_returncode can detect when
 # run_subprocess has been monkey-patched by a test (old mock pattern).
@@ -154,11 +166,12 @@ _BROAD_GENERATED_SCAN_DIR_NAMES = {
     "target",
     "venv",
 }
-# Single source of truth: `io/directory_scanner.py` (item #154) -- keeps this file and
-# `cli/main.py`'s equivalent guard from drifting out of sync. perf/#48: no longer aliased at
-# module level -- `_path_has_project_marker` / `_search_paths_include_workspace_root` below
-# import these constants function-locally (fast-path-unused; see the deferred-import note above
-# the `tensor_grep.cli.subprocess_policy` import).
+# Single source of truth: `io/scan_limits.py` (item #154; moved from `io/directory_scanner.py`
+# in campaign #6 / F2.4 -- `directory_scanner.py` still re-exports the same names for back-compat)
+# -- keeps this file and `cli/main.py`'s equivalent guard from drifting out of sync. perf/#48: no
+# longer aliased at module level -- `_path_has_project_marker` / `_search_paths_include_workspace_
+# root` below import these constants function-locally (fast-path-unused; see the deferred-import
+# note above the `tensor_grep.cli.subprocess_policy` import).
 _SEARCH_PATTERN_FLAGS = {"-e", "--regexp"}
 _SEARCH_LITERAL_FLAGS = {"-F", "--fixed-strings"}
 _SEARCH_PCRE2_FLAGS = {"-P", "--pcre2"}
@@ -635,7 +648,7 @@ def _search_args_paths_defaulted(search_args: list[str]) -> bool:
 
 
 def _path_has_project_marker(path: Path) -> bool:
-    from tensor_grep.io.directory_scanner import BROAD_WORKSPACE_PROJECT_MARKERS
+    from tensor_grep.io.scan_limits import BROAD_WORKSPACE_PROJECT_MARKERS
 
     for marker in BROAD_WORKSPACE_PROJECT_MARKERS:
         try:
@@ -671,7 +684,7 @@ def _search_paths_include_generated_root(paths: list[str]) -> bool:
 
 
 def _search_paths_include_workspace_root(paths: list[str]) -> bool:
-    from tensor_grep.io.directory_scanner import (
+    from tensor_grep.io.scan_limits import (
         BROAD_WORKSPACE_MARKED_ROOT_CHILD_THRESHOLD,
         BROAD_WORKSPACE_PROJECT_CHILD_THRESHOLD,
     )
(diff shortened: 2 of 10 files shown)


oimiragieo:
**Independent Opus gate: SHIP-WITH-NITS → NIT folded, now clean SHIP.**

An independent Opus review (framed to falsify the build agent's "all green" hypothesis) verified all six falsification targets on this fail-closed security surface — the relocated constants back the #702 unscoped-search refuse:

1. **Constant parity (byte-exact) — PASS.** AST-exec comparison of origin/main `directory_scanner.py` vs the new `scan_limits.py` returned zero membership diffs on every set: `_GENERATED_DIR_NAMES` (12), `_ALL_HEAVY_ROOT_DIR_NAMES` (4), `UNBOUNDED_VENDORED_ROOT_DIR_NAMES` (3 = `{external_repos, third_party, vendor}`), `BROAD_WORKSPACE_PROJECT_MARKERS` (10), and the three scalars (`3`, `8`, `1500`) — all identical.
2. **Re-export completeness — PASS.** `directory_scanner.py:21-38` re-exports all 5 public constants; every origin/main `from directory_scanner import <const>` resolves via re-export or was repointed. The `DirectoryScanner` class stayed put, so its importers are untouched.
3. **Consumer repoint correctness — PASS.** `main.py:46-52`, bootstrap's 3 guards, `scan_guardrails.py:7` → `scan_limits`; the 4th helper (`bootstrap.py:806`, needs `DirectoryScanner`) deliberately keeps the re-exported ceiling. No site missed, no value switched.
4. **Guard behavior invariance — PASS.** `git diff origin/main..HEAD` filtered of comments+imports yields **zero** code-logic lines in bootstrap.py/scan_guardrails.py — only the import *source* changed, not the refuse logic.
5. **Import-deferral safety (F2.5/F2.6) — PASS.** Function-local `json`/`shutil` present at all 3 sites (no NameError path); `-> object` resolves clean; the PEP-563 rationale (defers evaluation, not name-resolution) is sound.
6. **Bidirectional test validity — PASS (1 NIT, now folded).** The sync-pin `is`-identity test has real teeth; the fast-path probes flip True if a heavy import creeps back.

**NIT folded (commit `80a717a`):** `test_cli_package_init_has_no_eager_typing_import` asserted `"import typing" not in source`, which misses `from typing import Any` (that text contains "typing import", not "import typing") — the exact form F2.6 removed. Replaced with an AST parse of top-level statements rejecting **both** import forms while ignoring nested/`TYPE_CHECKING` imports. Bidirectionally validated (empty on the clean source, catches both forms). 13/13 pass; ruff/format clean.

Gate test run (CPU-safe, 2 changed files, pre-built venv, `tensor_grep.__file__` confirmed resolving into the worktree): **107 passed, 1 expected-skip** (native binary not built), 0 failed. No cargo/rustc; e2e routing-parity not run.

No MUST-FIX outstanding. Marking ready for review; enters the one-per-publish drain.

PR #712, opened 2026-07-22
test(retrieval): measure the shipped dense-weight config + ndcg regression floor

oimiragieo:
## Summary

Implements RANKED-QUEUE ITEM #7 of the +10% campaign: the shipped v1.93.2 dense-weight flip
(#191/#634, NL ndcg@10 0.3047->0.4466) had ZERO regression protection. This PR closes that gap
with benchmarks/tests-only infra -- no production source touched.

- **Thread `dense_weight`** through `run_rrf_arm`/`run_rrf_maxsim_arm` in
  `benchmarks/eval_late_rerank_quality.py` (byte-identical no-op at the default 1.0 -- every
  pre-existing call site is unaffected). Adds a fresh local `SHIPPED_DENSE_WEIGHT = 5.0` constant
  mirroring `cli/main.py`'s private `_FIND_DENSE_WEIGHT_ADAPTIVE_DEFAULT` (not imported, per the
  same "fresh local constant" precedent `DEFAULT_POOL_K` already sets for `reranker.py`'s private
  pool-size constant -- avoids dragging the ~17k-line CLI module into a benchmarks script).
- **New `rrf_shipped` arm** in `build_report`: the same bm25+dense RRF fusion as `rrf`, weighted to
  match what `tg find` actually ships. `rrf` itself is untouched (kept as the old 1:1 comparison
  baseline, clearly labeled).
- **New opt-in regression floor**: `tests/eval/test_retrieval_quality_regression.py`, `eval`
  marker (excluded from the default `-m "not eval"` CI sweep -- same discipline as
  `tests/eval/test_agent_accuracy.py`), asserts `ndcg@10 >= 0.40` on the committed 40-query golden
  set at the shipped weight.
- **Fixed the stale skip message**: `find`/`find+stack` used to claim "tg find pipeline not built
  yet" -- false since v1.77.0 (`tg find`'s CLI shipped that release, MCP followed in v1.78.0; see
  docs/BACKLOG.md #189). Renamed `SKIP_AWAITING_WAVE2` to `SKIP_FIND_ARM_NOT_WIRED` with an honest
  message: the pipeline exists, this harness just hasn't been wired to invoke it end-to-end yet.
- Hardened 2 pre-existing tests in `tests/unit/test_eval_late_rerank_quality.py` that silently
  assumed the ambient environment lacks `model2vec` (discovered while building this PR in a
  semantic-enabled venv) with explicit `dense_reason_override=...`, and added 4 new tests pinning
[...]

PR #716, opened 2026-07-23
perf(imports): stdlib resolver fast-path

oimiragieo:
## opt10 campaign #4: `tg imports` speed bundle

CPU-safe, results-identical performance change to the Python import resolver shared by
`tg imports` / `tg importers` / `tg agent` / `tg blast-radius` / `tg callers`. **Draft — an
independent Opus gate must run before merge (see "Load-bearing surface" below); do not un-draft.**

### Verify-plan-against-code (done before writing any code)

Re-verified every citation from the deep-dive against the real code on this worktree
(`origin/main` @ `b7d4725`, v1.93.3 + #712) before touching anything:

| Claim | Verified against real code |
|---|---|
| `_python_module_candidates` at repo_map.py:6454 | **Confirmed** (pre-edit). Signature: `(importer_path: Path, module_name: str, repo_root: Path \| str \| None = None, *, level: int = 0) -> dict[str, Any]`. |
| Static-import walk at repo_map.py:~5994 | **Confirmed** — `_python_imports_with_lines` (repo_map.py:5971, pre-edit), `for node in ast.walk(tree):` at line 5994. |
| `_python_dynamic_import_entries` at repo_map.py:1726, own walk at :1761 | **Confirmed exactly** — `for node in ast.walk(tree):` at line 1762 (pre-edit). |
| Callers of `_python_dynamic_import_entries` | **Enumerated, exactly 2**, both inside `repo_map.py`: `_python_imports_and_symbols` (repo_map.py:1911, call at line 2010 — the reverse-import alias-graph prefilter feeding `tg importers`/`tg blast-radius`/`tg callers`) and `_python_imports_with_lines` (repo_map.py:5971, call at line 6039 — the `tg imports` line-precise primitive, F4.2's target). No callers outside `repo_map.py` (grepped the whole repo; only `tests/unit/test_file_deps.py` and `CHANGELOG.md` reference the name). |
| Return shape for an external/unresolved module | **Read, not assumed.** `_python_module_candidates`'s `paths` list is populated with normalized-but-NOT-existence-checked candidate `Path`s (via `_resolved_path_str` = `Path.resolve()`, no `.exists()`/`.is_file()` call inside this function at all) — existence is checked by the CALLERS: `_resolve_raw_import_entry` (repo_map.py:16197-16212, pre-edit) does `Path(current).is_file()` per candidate to set `resolved`/`external`; `_python_module_match_details` (repo_map.py:6516-6550) does an `any(str(candidate) == resolved_definition ...)` membership check. For `level == 0`, `provenance`/`confidence` are set **unconditionally** to `["python-path-heuristic"]` / `0.7` before any candidate is even probed (pre-edit lines 6495-6496) — i.e. those two fields are IDENTICAL whether 0 or 12 candidates get constructed. This is the exact shape the fast-path must reproduce. |
| Local-module-shadow detection mechanism | **Read.** There is no separate "is this local" branch — the general path just builds `root.joinpath(*parts).with_suffix(".py")` / `root.joinpath(*parts, "__init__.py")` for every root in `_python_candidate_roots` and lets the caller's `.is_file()` decide. So the fast-path's shadow gate must independently probe the SAME two candidate shapes over the SAME roots before skipping construction — see "Shadow-safety proof" below. |

The deep-dive's mechanism matched the real code closely; the one adjustment made was **scoping
the fast-path to `len(parts) == 1`** (a bare `import json`, not a dotted `import os.path`) —
the deep-dive's sketch didn't specify this, and a dotted access needs a different shadow-probe
shape (`root/parts[0]` as a directory, not `root/parts[0].py`) that isn't worth the extra
complexity given single-segment imports are already the dominant shape.

### What shipped

**F4.3 (primary) — shipped.** `_python_module_candidates` (repo_map.py) fast-paths
`level == 0, len(parts) == 1, parts[0] in sys.stdlib_module_names` past the general multi-root
candidate construction (which does ~2 `Path` builds x ~5-6 roots, each pushed through
`_resolved_path_str`'s resolve-and-dedupe), returning `{"paths": [], "provenance":
["python-path-heuristic"], "confidence": 0.7, "path_provenance": {}}` — the exact shape the
general path produces for ANY level==0 import when nothing resolves.

**F4.2 (bonus) — shipped, provably clean.** Rather than touch `_python_dynamic_import_entries`
externally (its OTHER caller, `_python_imports_and_symbols`, must see zero behavior change), its
loop body was extracted unchanged into a new pure per-node helper,
`_python_dynamic_import_entry_for_call(node) -> dict | None`. `_python_dynamic_import_entries`
now just does `for node in ast.walk(tree): entry = _python_dynamic_import_entry_for_call(node)`
— same walk, same call-once-per-node, same output, zero behavior change for its existing caller.
`_python_imports_with_lines` then folds a call to that SAME helper into its own existing
`ast.walk(tree)` loop (an `elif isinstance(node, ast.Call):` branch alongside the existing
`Import`/`ImportFrom` branches), accumulating into a separate `dynamic_entries` list and
concatenating `entries + dynamic_entries` at the end — instead of calling
`_python_dynamic_import_entries(tree)` wholesale (a second full-tree walk).

### Shadow-safety proof (F4.3's load-bearing correctness risk)

**Order-preservation is provable, not incidental:** `ast.walk` is a deterministic BFS (a
`collections.deque`, pop-left / extend-with-children — no randomization) over a fixed tree, so
filtering it ONCE for two disjoint node-type predicates (`Import`/`ImportFrom` vs `Call`) and
concatenating the two per-predicate result lists produces the IDENTICAL two orderings that
filtering it TWICE (the old shape) would produce. `ast.Import`/`ast.ImportFrom` and `ast.Call`
are disjoint node types, so no entry can be double-counted or reordered relative to its own
kind.

**TDD proof it's load-bearing, not decorative:** per the task instructions, I temporarily
stripped the shadow-probe loop down to an unconditional
`if level == 0 and len(parts) == 1 and parts[0] in sys.stdlib_module_names: return {"paths": [],
...}` (no gate) and re-ran the two shadow tests:

```
tests/unit/test_file_deps.py::test_build_file_imports_stdlib_shadowed_by_local_module_resolves_to_local_file
  FAILED - assert entry["external"] is False  ->  AssertionError: assert True is False
tests/unit/test_file_deps.py::test_build_file_imports_stdlib_shadowed_by_local_package_resolves_to_local_file
  FAILED - assert entry["external"] is False  ->  AssertionError: assert True is False
```

Then restored the shadow-gated version — both pass again (see test list below). The gate
(`_python_candidate_roots` + a cheap `.is_file()`/`.is_dir()` probe per root, falling CLOSED to
the general path on any doubt, including `OSError`) is what makes them pass.

### Results-identical golden method

Captured `build_file_imports` output for a mixed fixture (`import os`, `import sys`, `import
json`, `import collections.abc`, `import numpy`, `import pkg.helpers`, `from . import helpers`)
against the **unmodified pre-change code**, then re-ran the identical fixture after the change —
every field (`module`/`line`/`resolved`/`external`/`provenance`/`resolution_confidence`) came
back byte-identical:

```
BEFORE (unmodified):
{"external": true,  "line": 1, "module": "os",              "provenance": ["python-path-heuristic"], "resolution_confidence": 0.0, "resolved": null}
{"external": true,  "line": 2, "module": "sys",             "provenance": ["python-path-heuristic"], "resolution_confidence": 0.0, "resolved": null}
{"external": true,  "line": 3, "module": "json",            "provenance": ["python-path-heuristic"], "resolution_confidence": 0.0, "resolved": null}
{"external": true,  "line": 4, "module": "collections.abc", "provenance": ["python-path-heuristic"], "resolution_confidence": 0.0, "resolved": null}
{"external": true,  "line": 5, "module": "numpy",           "provenance": ["python-path-heuristic"], "resolution_confidence": 0.0, "resolved": null}
{"external": false, "line": 6, "module": "pkg.helpers",     "provenance": ["python-path-heuristic"], "resolution_confidence": 0.7, "resolved": ".../project/pkg/helpers.py"}
{"external": false, "line": 7, "module": "helpers",         "provenance": ["relative"],              "resolution_confidence": 1.0, "resolved": ".../project/pkg/helpers.py"}

AFTER (this PR): identical, field-for-field.
```

The raw `_python_module_candidates()` dict DOES differ internally for the 3 bare-stdlib entries
(`paths: []` now vs. a populated-but-entirely-nonexistent 6-candidate list before) — this is
INTENTIONAL and provably inert: neither real consumer (`_resolve_raw_import_entry`'s
`.is_file()` filter, `_python_module_match_details`'s membership check) can distinguish "empty
list" from "list of only-nonexistent paths", since neither contains a real file. The shadow case
(`import json` with a local `json.py`) is the one case where this WOULD matter, and it's proven
identical above (`test_build_file_imports_stdlib_shadowed_by_local_module_resolves_to_local_file`).

This test is now baked into the suite as
`test_build_file_imports_mixed_imports_results_identical_to_pre_fastpath_baseline`.

### Informal timing sanity-check (NOT a certified benchmark-harness run)

Small, bounded, in-process microbenchmarks — evidence the two fast-paths do something
measurable, not a claim run through the project's official benchmark harness:

- **F4.3**: `_python_module_candidates("os", ...)` over 1200 distinct importer paths (fresh
  `_resolved_path_str` cache per path, fast/slow interleaved per-path to cancel ordering bias):
  **~2.7x faster** (1494 us/call fast-path-on vs 4095 us/call forced-off).
- **F4.2**: isolated walk-only comparison on a pre-parsed in-memory AST (excludes file I/O and
  AST-parse-cache noise; 5 alternating repetitions x 20000 calls to cancel warmup bias): merged
  single-walk **~1.75x faster** than the reconstructed old two-walk shape (0.62s median vs
  1.09s median per rep).
- A naive first attempt at this timing (same-triple repeated calls, sequential non-alternating
  loops) produced misleading numbers in BOTH directions — once from `_resolved_path_str`'s LRU
  cache trivializing repeat "slow path" calls, once from an OS file-content-cache ordering bias
  between two sequential loops reading the same 1500 files. Both are noted here as "profile the
  actual thing, not a benchmark artifact" receipts, not hidden.

### Test list (targeted pytest, PYTHONPATH into this worktree's src/)

`C:/dev/projects/tensor-grep/.venv/Scripts/python.exe -m pytest tests/unit/test_file_deps.py -p no:cacheprovider -q`
— **95 passed** (87 pre-existing + 8 new):

- `test_python_module_candidates_stdlib_fastpath_returns_general_path_shape` (new) — direct
  dict-shape contract for `os`/`sys`/`json`.
- `test_python_module_candidates_dotted_stdlib_access_does_not_take_fastpath` (new) — proves the
  `len(parts) == 1` gate via `collections.abc` still populating candidates.
- `test_build_file_imports_stdlib_shadowed_by_local_module_resolves_to_local_file` (new) — **the
  load-bearing shadow test** (a local `json.py` at repo root).
- `test_build_file_imports_stdlib_shadowed_by_local_package_resolves_to_local_file` (new) — the
  package-shadow sibling (a local `queue/__init__.py`).
- `test_build_file_imports_relative_import_of_stdlib_named_sibling_still_resolves` (new) —
  proves `level > 0` never takes the fast path.
- `test_build_file_imports_mixed_imports_results_identical_to_pre_fastpath_baseline` (new) — the
  golden byte-identical test.
- `test_python_imports_with_lines_merges_dynamic_walk_into_single_ast_walk_pass` (new) — F4.2
  walk-count (`monkeypatch`'d `ast.walk` counter, asserts exactly 1 call) + order-identity proof.
- `test_python_dynamic_import_entries_unchanged_after_extraction` (new) — regression guard for
  `_python_dynamic_import_entries`'s OTHER caller (`_python_imports_and_symbols`).

Also re-ran for collateral-impact coverage on functions that consume `_python_module_candidates`
/ `_python_imports_with_lines` transitively (`build_repo_map`, `build_agent_capsule`,
caller/blast-radius scans): `test_repo_map_targets.py`, `test_repo_map_spans.py`,
`test_repo_map_deadline.py`, `test_ast_parse_cache.py`, `test_medlow2_repo_map.py`,
`test_codemap.py`, `test_agent_capsule_outbound_deps.py`,
`test_agent_reverse_import_graph_deadline_222.py`, `test_impact_callers_shared_map.py`,
`test_graph_completeness_oracle.py` — **225 passed**. Plus the `agent_capsule_*` /
`agent_vendored_subtree_*` / `agent_suggested_scope` / `agent_ignore` group — **103 passed**.
**423 tests total, 0 failures.**

Gates: `ruff check` and `ruff format --check --preview` both clean on the 2 changed files
(`src/tensor_grep/cli/repo_map.py`, `tests/unit/test_file_deps.py`).

Verified `python -c "import tensor_grep; print(tensor_grep.__file__)"` resolves into this
worktree's `src/` before trusting any RED/GREEN (stale-venv trap).

### CPU-SAFE compliance

No `cargo`/`rustc`/`maturin` run. No `tests/e2e/test_routing_parity.py` run (verified the Rust
front-door is untouched by READING only — this change is 100% Python, `repo_map.py` +
`test_file_deps.py`). All test/format/lint runs were targeted, bounded, pre-built-venv Python.

### Disjoint from PR #713

Confirmed `_truncate_source_text_to_budget` (PR #713's territory) lives at repo_map.py:8923-9062
(pre-edit) — this PR's edits are at lines ~1726-1800 and ~5991-6055 and ~6454-6549 (pre-edit
numbering), nowhere near #713's range.

### Load-bearing surface: flag for independent review

**F4.3's shadow gate** (`_python_module_candidates`'s `if level == 0 and len(parts) == 1 and
parts[0] in sys.stdlib_module_names:` block) is the one piece of this PR with real correctness
risk — a wrong shadow probe silently mis-resolves any repo that happens to ship a top-level
module/package named the same as a stdlib module (`json.py`, `queue.py`, `types.py`, `token.py`,
`copy.py`, `string.py`, `pdb.py`, etc. are all plausible real filenames). Please independently
re-derive: (a) that `_python_candidate_roots` is genuinely the same root set the general path
uses (not a subset), and (b) that `(root / f"{name}.py").is_file() or (root / name).is_dir()`
can never be `False` while a real candidate the general path WOULD have matched exists at that
root. This PR is left in draft for that gate — do not merge without it.

### DIFF
--- a/src/tensor_grep/cli/repo_map.py
+++ b/src/tensor_grep/cli/repo_map.py
@@ -7,6 +7,7 @@ import json
 import math
 import os
 import re
+import sys
 import tempfile
 import threading
 import time
@@ -1759,28 +1760,43 @@ def _python_dynamic_import_entries(tree: ast.AST) -> list[dict[str, Any]]:
     """
     entries: list[dict[str, Any]] = []
     for node in ast.walk(tree):
-        if not isinstance(node, ast.Call) or not _is_python_dynamic_import_call(node):
-            continue
-        literal_module: str | None = None
-        if (
-            node.args
-            and isinstance(node.args[0], ast.Constant)
-            and isinstance(node.args[0].value, str)
-        ):
-            literal_module = node.args[0].value
-        dynamic_unresolved = literal_module is None
-        if literal_module is not None and (
-            literal_module.startswith(".") or _python_dynamic_import_call_is_relative(node)
-        ):
-            dynamic_unresolved = True
-        entries.append({
-            "module": literal_module or "",
-            "line": int(node.lineno),
-            "level": 0,
-            "dynamic": True,
-            "dynamic_unresolved": dynamic_unresolved,
-        })
+        entry = _python_dynamic_import_entry_for_call(node)
+        if entry is not None:
+            entries.append(entry)
     return entries
+
+
+def _python_dynamic_import_entry_for_call(node: ast.AST) -> dict[str, Any] | None:
+    """Per-node half of `_python_dynamic_import_entries` above: given a single AST node, return
+    its dynamic-import entry dict if `node` is one of the 3 dynamic-import call shapes (see
+    `_is_python_dynamic_import_call`), else `None`. Pulled out of that function's loop body
+    unchanged (same literal-extraction, same relative-literal fail-closed check, same entry
+    shape) so a caller that ALREADY walks `tree` for another purpose can fold this per-node check
+    into its own walk instead of paying for a second whole-tree `ast.walk` -- see
+    `_python_imports_with_lines` (opt10 F4.2), which merges its static-import walk with this one.
+
+    `_python_dynamic_import_entries` itself keeps doing its own `ast.walk` and calling this once
+    per node exactly as its inlined loop body used to -- this extraction changes nothing about
+    its behavior, signature, or the entries it returns for its existing caller
+    (`_python_imports_and_symbols`'s reverse-import alias-graph prefilter).
+    """
+    if not isinstance(node, ast.Call) or not _is_python_dynamic_import_call(node):
+        return None
+    literal_module: str | None = None
+    if node.args and isinstance(node.args[0], ast.Constant) and isinstance(node.args[0].value, str):
+        literal_module = node.args[0].value
+    dynamic_unresolved = literal_module is None
+    if literal_module is not None and (
+        literal_module.startswith(".") or _python_dynamic_import_call_is_relative(node)
+    ):
+        dynamic_unresolved = True
+    return {
+        "module": literal_module or "",
+        "line": int(node.lineno),
+        "level": 0,
+        "dynamic": True,
+        "dynamic_unresolved": dynamic_unresolved,
+    }
 
 
 # ---------------------------------------------------------------------------
@@ -5983,14 +5999,28 @@ def _python_imports_with_lines(path: Path) -> list[dict[str, Any]]:
         return []
 
     entries: list[dict[str, Any]] = []
+    dynamic_entries: list[dict[str, Any]] = []
     # Nested-scope recall fix: `ast.walk` (not `tree.body`) so a plain `import`/`from ... import`
     # STATEMENT nested inside a function body, an `if`/`try` block, or an `if TYPE_CHECKING:`
     # guard is collected too -- `tree.body` only ever visited module-top-level statements,
     # silently missing anything scope-nested (a `tg imports`/`tg importers` recall gap;
-    # `result_incomplete` stayed False, so the omission was invisible). This mirrors
-    # `_python_dynamic_import_entries` below, which already whole-tree-walks for dynamic
-    # `__import__`/`import_module(...)` CALLS; this closes the same gap for static import
-    # statements, the shape `tg imports`/`tg importers` mostly see.
+    # `result_incomplete` stayed False, so the omission was invisible).
+    #
+    # opt10 F4.2 speed fix: this single walk ALSO picks up `__import__`/`import_module`/
+    # `importlib.import_module` CALLS -- the #93 SUB-1 dynamic-import shape -- via
+    # `_python_dynamic_import_entry_for_call` (the extracted per-node half of
+    # `_python_dynamic_import_entries`), instead of a SEPARATE second `ast.walk(tree)` over the
+    # same tree the way this used to call that function wholesale. `ast.Import`/`ast.ImportFrom`
+    # and `ast.Call` are disjoint node types, so folding both checks into one walk and
+    # accumulating into two separate lists (`entries` for static, `dynamic_entries` for dynamic)
+    # produces the IDENTICAL two per-kind orderings `ast.walk` would produce run separately --
+    # `ast.walk` is a deterministic traversal of a fixed tree, so filtering it once for two
+    # disjoint predicates and concatenating the two result lists (`entries + dynamic_entries`,
+    # same order as the old `entries.extend(_python_dynamic_import_entries(tree))`) is exactly
+    # equivalent to filtering it twice. See
+    # test_python_imports_with_lines_merges_dynamic_walk_into_single_ast_walk_pass (walk-count +
+    # order-identity proof) and `_python_dynamic_import_entries`, which is UNCHANGED and still
+    # walks `tree` on its own for its other caller (`_python_imports_and_symbols`).
     for node in ast.walk(tree):
         if isinstance(node, ast.Import):
             for alias in node.names:
@@ -6012,11 +6042,11 @@ def _python_imports_with_lines(path: Path) -> list[dict[str, Any]]:
                         "line": int(node.lineno),
                         "level": int(node.level),
                     })
-    # #93 SUB-1: `ast.walk` (not `tree.body`) picks up `__import__`/`import_module`/
-    # `importlib.import_module` calls anywhere -- inside a function, a conditional, a
-    # try/except -- that the static scan above (now itself an `ast.walk`, see above) never
-    # visits, since these are `ast.Call` expressions, a different node type entirely.
-    entries.extend(_python_dynamic_import_entries(tree))
+        elif isinstance(node, ast.Call):
+            dynamic_entry = _python_dynamic_import_entry_for_call(node)
+            if dynamic_entry is not None:
+                dynamic_entries.append(dynamic_entry)
+    entries.extend(dynamic_entries)
     return entries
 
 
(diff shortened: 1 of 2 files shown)


oimiragieo:
**Independent Opus gate: SHIP** (zero MUST-FIX, 2 nits banked).

An independent Opus review (framed to falsify the build agent's "results-identical" hypothesis) verified all six targets with BEFORE/AFTER execution against the PR-base blob — every downstream output of `tg imports`/`importers`/blast-radius is **byte-identical**, empirically proven:

1. **F4.3 shadow-safety — PASS.** The fast-path iterates the *identical* `_python_candidate_roots(...)` the general path uses; its shadow probe (`<root>/<name>.py` is_file OR `<root>/<name>` is_dir) is a strict **superset** of the general path's local-shadow check, so it catches every shadow and only ever over-fires (dir-without-`__init__` → falls to slow path → still external). A local `json.py` at repo root / package / importer-dir / sys.path-hack root all resolve identically before/after.
2. **F4.3 shape — PASS (nuance → nit 1).** The only dict delta is `paths: []` vs a list of ~6 **non-existent** candidate paths; both consumers touch `paths` only via existence checks against real files → identical `external=True, confidence=0.0`. Zero classification drift.
3. **F4.2 walk-merge — PASS.** `_python_dynamic_import_entry_for_call` is a byte-identical extraction; `ast.Import`/`ImportFrom`/`Call` are disjoint so one filtered walk + concat == two walks, same order. `_python_dynamic_import_entries` unchanged for its other caller.
4. **Golden not a tautology — PASS.** Ran the golden under genuine BEFORE code; the hardcoded `expected` == real BEFORE output (not after-vs-after).
5. **Bidirectional teeth — PASS.** Stripping the gate flips both shadow tests RED; intact → GREEN. (The reviewer caught+bypassed a stale `.pyc` + a `conftest sys.path.insert` during verification — final run airtight.)
6. **Edge cases — PASS.** `os.path`/`collections.abc` (`len(parts)==1` guard → general path); no `stdlib_module_names` version hazard (membership gates only *which path*, not the result — stdlib is reported as external either way); `level>0` relatives skip the fast-path and resolve to local siblings.

`test_file_deps.py`: 95 passed / 0 failed via the sanctioned venv (`tensor_grep.__file__` confirmed in-worktree); no cargo/maturin/e2e.

**Nits BANKED (no code change):** (1) soften the "byte-identical value" framing to "observationally identical" (the `paths` field differs but is invisible downstream; the in-code comment at repo_map.py:6527-6530 already states this correctly) — I'll carry the accurate wording into the squash-merge message; (2) optionally add a local-shadow row to the results-identical golden (shadow is already covered by two dedicated teeth-verified tests).

**Merge-hygiene:** this PR's base trails current `origin/main` and repo_map.py is a wave-2 conflict point — the 87-line delta doesn't touch the reviewed functions so the rebase should be clean, but I'll re-verify the reviewed regions are intact at merge time. Marking ready for review; enters the one-per-publish drain.

PR #717, opened 2026-07-23
perf(retrieval): max-combine fusion — accuracy ndcg@10 +62.6%

oimiragieo:
## Summary

`opt10` Phase-2 ACCURACY lever: switch retrieval fusion from sum-RRF to **max-combine** (best-rank-wins), default-flipped with zero call-site edits — **plus a query-adaptive routing fix (commit 2) that keeps single-token literal/identifier `tg find` queries on the sum path**, after an independent Opus gate falsified this PR's original "never worse than sum" claim on the committed `literal_golden.jsonl` slice.

**LOAD-BEARING shared-ranking default change — independent Opus gate ran once already (SHIP-WITH-NITS, folded below); re-verifying the fold before un-draft.**

## UPDATE (commit 2): the gate found a real regression, now fixed

The gate confirmed every correctness claim below (rrf arm +62.6% reproduced at both endpoints, capsule 16/16, the tie mechanism mutation-proven) **but correctly falsified** this PR's original "never worse than sum / 6/6" framing: on the committed `benchmarks/datasets/literal_golden.jsonl` slice (10 single-token literal/identifier queries, `dense_weight=1.0` as `_find_dense_weight` already protects them at), max-combine is a **real ndcg@10 regression vs sum: 0.9631 vs 1.0000 (−0.0369)**. `identifier3_golden.jsonl` (10 longer, 3+-morpheme identifiers) is a wash — 1.0000 both modes. The **6/6 dogfood queries and the 40-query golden set below are ALL natural-language** — they do not cover, and never claimed to cover, the literal/identifier query shape where this regression lives. That was this PR's honesty gap: the "never worse than sum" line implied a claim broader than what was actually tested.

**Root cause:** a literal query's true answer is frequently ranked #1 by BOTH the bm25 leg and the dense leg independently. Under `combine="max"`, that doc's fused score is just its best SINGLE-leg term (`1/(k+1)`) — identical to a false competitor that is rank-1 in only ONE leg. The two tie bit-for-bit, and the ascending-chunk-index tie-break can pick the wrong one. Under `combine="sum"`, the true answer's second-leg agreement adds a small but real bonus that correctly breaks the tie — the exact signal `max` discards.

**The fold (not just documented — fixed):** `rank_chunks` (`reranker.py:275`, now `:291` post-edit) gains `combine: Literal["sum", "max"] = "max"` (matching `reciprocal_rank_fusion`'s own default, so an omitted kwarg — `rerank_hybrid`'s `--semantic` path — is unaffected). `tg find`'s two `rank_chunks` call sites (`_execute_find`, `main.py`) now also pass `combine=_find_combine_mode(query)` — a new sibling of `_find_dense_weight` sharing the exact same whitespace predicate (`_find_is_single_token_query`, factored out of both so the two adaptive knobs can never disagree): a single whitespace-free token → `"sum"` (byte-identical to the pre-max-flip behavior, recovering the regression); genuinely multi-word NL → `"max"` (keeps the +62.6% win). `reciprocal_rank_fusion`'s own DEFAULT is **untouched** — the scorecard's `rrf` arm still scores exactly 0.4953.

### Verified after the fold

| check | result |
|---|---|
| eval `rrf` arm (default max, 40 NL, direct `reciprocal_rank_fusion` call — untouched by this fix) | **0.4953, unchanged** |
| `literal_golden.jsonl` via `rank_chunks(dense_weight=1.0, combine=...)` — the exact shipped `tg find` path for every one of its 10 real queries | `combine="max"`: **0.9631** (byte-matches the gate's finding exactly, incl. the single missing query `lit-02`/"remember") → `combine="sum"` (what `_find_combine_mode` now actually returns for all 10): **1.0000 — regression fully recovered** |
| `identifier3_golden.jsonl` | 1.0000 both modes — unaffected, as expected (a wash) |
| 40-query NL set via `rank_chunks(dense_weight=5.0, combine=...)` — the exact shipped `tg find` path for every one of its queries (`_find_combine_mode` confirmed to return `"max"` for all 40) | **0.6026 — the dense-alone ceiling, unchanged** |
| Capsule agent-accuracy gate (`tests/eval/test_agent_accuracy.py`) | **16/16, unchanged** |
| Routing confirmed directly: `_find_combine_mode(q)` for every real query in both golden slices | `"sum"` for all 10+10; `"max"` for all 40 NL queries |
| New tests | `test_rank_chunks_combine_parameter_threads_to_fusion` (wiring spy), `test_rank_chunks_combine_sum_recovers_literal_regression_scenario` (a hand-built, real-`Bm25Index`/`DenseIndex` reproduction of the exact tie mechanism — proves `max_order[0]==0` (wrong) vs `sum_order[0]==1` (correct) deterministically), `test_find_combine_mode_helper` (direct-call, mirrors the existing dense_weight identifier list), `test_find_combine_mode_routes_single_token_to_sum_multiword_to_max_end_to_end` (CLI-level), `test_find_combine_mode_dense_unavailable_retry_also_routes` (the F1 BM25-only retry path also routes correctly) |
| `ruff check` / `ruff format --check --preview` / `mypy` on all changed files | clean |
| Full downstream suite | **128/128** (`test_retrieval_fusion` 21, `test_rank_chunks` 9, `test_reranker` 18, `test_reranker_hybrid` 14, `test_eval_late_rerank_quality` 20, `test_find_command` 24, `test_search_semantic_rerank` 5, `test_semantic_search_flag` 18 incl. real dense model, `test_bm25_search_flag` 3 — some totals differ slightly from commit 1's report after the spy-signature fix below) |

One incidental fix needed to land this: a pre-existing test (`test_rank_chunks_dense_weight`) had a local spy with a hardcoded `reciprocal_rank_fusion(rankings, *, k, weights)` signature that didn't accept the new `combine` kwarg `rank_chunks` now always passes — updated its signature to accept and forward `combine`, no behavior change to what it asserts.

**Methodology note (transparency):** I initially tried to reproduce the `literal_golden.jsonl` regression by running the *full* `_execute_find` pipeline (its own independent repo walk + chunking) and got 1.0 for both `max` and `sum` — it did not reproduce the gate's number. Root cause: `_execute_find`'s own file-walk order assigns different chunk INDICES than the eval harness's `load_corpus_files`+`chunk_file` convention, and the tie-break is by ascending chunk index — a different index assignment can accidentally dodge the tie for this one query. I do not rely on that walk-order-dependent luck: the verification above uses `rank_chunks` directly over the harness's own chunk build (the same convention the frozen 40-query numbers already use), which is the apples-to-apples comparison and is what byte-matches the gate's cited 0.9631. The routing fix is unconditional on query SHAPE (whitespace), not on walk order, so it holds regardless of which convention chunks it.

---

## The change (file:line)

`src/tensor_grep/core/retrieval_fusion.py`, `reciprocal_rank_fusion` (`:18`-`:56` pre-change, core loop was `:53`). Added `combine: Literal["sum", "max"] = "max"`:

- `combine="max"` (**new default**): `scores[i] = max(scores.get(i, 0.0), (1.0/(k+rank))*weight)` — a chunk's fused score is the single best leg contribution it earns, never the sum of every leg.
- `combine="sum"`: byte-identical to the pre-change behavior, for any call site that must keep it.

Every existing call site (`reranker.py`'s `rank_chunks` — used by both `tg search --semantic` and `tg find` — and the eval harness's `rrf`/`rrf_shipped` arms) calls `reciprocal_rank_fusion(...)` **without** `combine`, so the default change moves all of them with zero call-site edits.

## Why (mechanism)

A weak/near-floor leg (BM25 on a vocabulary-mismatched NL query) can now only ever **help** a chunk's rank — by ranking it even higher than the strong (dense) leg did — and can never **drag** the strong leg's pick down merely by failing to rank it or ranking it worse. Under sum, a doc that the dense leg loves but BM25 never sees is diluted by every other doc's tiny same-two-leg sum; under max, that dilution disappears. **Caveat (see UPDATE above): this same discarding-of-agreement is exactly what regresses a literal query where BOTH legs independently agree — the routing fix scopes `max` to where it actually helps.**

## Eval re-run (frozen golden-set harness, live, not replayed)

`benchmarks/eval_late_rerank_quality.py`, 40 queries (100% NL — see the UPDATE's caveat), oracle-validated, 2/2 runs byte-identical:

| arm | ndcg@10 before | ndcg@10 after | recall@10 | mrr |
|---|---|---|---|---|
| `bm25` | 0.1093 | **0.1093 (unchanged)** | 0.25 | 0.068 |
| `dense` | 0.6026 | **0.6026 (unchanged)** | 0.90 | 0.507 |
| `rrf` (the scored arm) | 0.3047 | **0.4953 (+62.6%)** | 0.55 → 0.825 | 0.228 → 0.391 |
| `rrf_shipped` (dense_weight=5.0, the real `tg find` production weight for NL queries) | *(not measured pre-change; sum+weighted independently reproduced at 0.4466)* | **0.6026 — reaches the dense-alone ceiling exactly** | 0.90 | 0.507 |
| `rrf+maxsim` | 0.0679 | 0.0743 | 0.225 | 0.032 |

`rrf+maxsim` stays clearly below bm25/rrf both before and after — already a known-dead arm per the eval script's own docstring ("twice-confirmed dead at the current model tier"), unaffected in verdict.

I independently reproduced the pre-change baseline byte-for-byte (rrf 0.3047, dense 0.6026 — exact match to the frozen `baseline_results.json` from the opt10 campaign) before trusting the after-change numbers.

Determinism: `--runs 2` → 2/2 byte-identical. Oracle: `--validate-oracle` → PASSED (gold=1.0 exact, reversed/empty at-or-below ceiling).

## Capsule agent-accuracy gate (hard 16/16 floor)

`pytest tests/eval/test_agent_accuracy.py` (real `tg prepare` subprocess, 16 golden tasks): **16/16, unchanged** (re-confirmed again after commit 2). Confirmed `agent_capsule.py` has zero `reciprocal_rank_fusion` references (AST retrieval, not dense/RRF) — it was structurally unaffected, and the live re-run confirms it, including the task literally named `"fix a weighting bug in reciprocal_rank_fusion between BM25 and dense scores"` resolving correctly to `core/retrieval_fusion.py`.

## `search --rank` decision: no scoping needed — it structurally never touches this function

Traced both flags before assuming the mandate's framing was accurate:

- `tg search --rank`/`--bm25` (no `--semantic`) → `main.py:8029-8032` → `rerank_by_bm25` (`reranker.py:161`). This function builds a **plain BM25 index with no dense leg at all** and never calls `reciprocal_rank_fusion`. It is **structurally unaffected** by this change — confirmed by 3/3 green in `tests/integration/test_bm25_search_flag.py` and 18/18 in `tests/unit/test_reranker.py`.
- `tg search --semantic` (`main.py:8010-8023` → `_apply_semantic_rerank` → `rerank_hybrid` → `rank_chunks` → `reciprocal_rank_fusion`) and `tg find` (`main.py`, direct `rank_chunks` call) are the **actual** consumers of RRF fusion on the search side. `--semantic` never passes `combine` (stays on plain `max`, unaffected by commit 2's fix, and its own test suite is fully green — 14/14 `test_reranker_hybrid.py`, 5/5 `test_search_semantic_rerank.py`, 18/18 `test_semantic_search_flag.py` including the real dense model). `tg find` is the one surface that now routes per-query (commit 2).

One real regression caught and fixed WITHIN commit 1: `test_chunk_absent_from_a_leg_contributes_zero` (pre-existing, `test_retrieval_fusion.py`) asserted a SUM-specific tie outcome that flips under the new default — fixed by pinning `combine="sum"` explicitly (its intent was always to test the sum path). Every other pre-existing test in that file is now pinned to `combine="sum"` too, as an explicit byte-identical regression guard.

## Weighted-max interaction (dense_weight=5.0, `tg find`'s real NL production weight)

Not just "not harmful" — **optimal for NL**: `rrf_shipped`/the routed multi-word find path (max-combine + the shipped `dense_weight=5.0`) scores **exactly** the dense-alone ceiling (0.6026). Mechanism: `max(bm25_term, 5.0 * dense_term)` — a 5x-amplified dense term all but always wins the max unless BM25 ranks the doc at least 5x higher, so the fused order collapses toward (and here, onto) the dense ranking. Under the OLD sum+weighted behavior (independently reproduced: 0.4466), the weak BM25 leg's contribution was still being *added in*, diluting the fused result relative to dense alone.

## Real-corpus dogfood (tensor-grep's own `src/`, 6 real queries — **all NL/multi-word, see the UPDATE's caveat**, `tg find`'s real pipeline `_execute_find`)

MAX vs SUM top-5, forced via the same monkeypatch pattern `test_rank_chunks_dense_weight` already uses (`reranker_module.reciprocal_rank_fusion`), so this exercises the shipped code path end-to-end, not a toy reproduction:

| query | top-1 (both modes) | notes |
|---|---|---|
| "where is the broad scan guard enforced" | `cli/scan_guardrails.py` (identical) | positions 2-5 reshuffle among plausible files (`apply_policy.py` vs `mcp_server.py`) |
| "how does the daemon warm cache work" | `cli/session_daemon.py` x3 (identical) | MAX concentrates MORE of top-5 on the correct file (4/5 vs 3/5 under sum) |
| "fusion of bm25 and dense rankings" | `core/reranker.py` (identical, full top-5 identical) | correctly surfaces the real fusion-consumer file at rank 1 |
| "where is the symlink follow disclosure prevented" | `cli/evidence_signing.py` (identical) | neither mode surfaces `io/directory_scanner.py` (the ideal answer) in top-5 — a **pre-existing** gap shared by both modes equally, not a MAX regression |
| "how does the backend fail closed contract work" | `cli/mcp_server.py` (identical) | genuinely a wash: sum surfaces more `backends/*.py` files, max surfaces `core/pipeline.py` (also a legitimate fail-closed-contract file per `AGENTS.md`) |
| "where is the session daemon idle shutdown timeout configured" | `cli/session_daemon.py` x5 (identical, full top-5 identical) | matches the capsule gate's own golden answer for this exact task phrasing |

**Corrected verdict (was overclaimed as "never worse than sum / 6/6" — see UPDATE):** on these 6 **NL** queries, MAX matches or is comparably-plausible to SUM at rank 1 in every case. This says nothing about literal/identifier queries, where a real regression existed and is now fixed by the routing above — that is the honesty gap the gate correctly caught.

## TDD

- `tests/unit/test_retrieval_fusion.py`: every pre-existing test now pins `combine="sum"` explicitly (the byte-identical regression guard for the original formulation). Added 6 new tests for `combine="max"`: default-is-max, the exact hand-computed fixture from the campaign brief, the max-specific tie/floor behavior, weights-apply-before-max, determinism, and invalid-`combine` raises. 21/21 pass.
- `tests/unit/test_rank_chunks.py`: +2 tests (combine-threading wiring, the hand-built tie-mechanism regression proof); 1 pre-existing test's local spy signature updated to accept the new `combine` kwarg. 9/9 pass.
- `tests/unit/test_find_command.py`: +3 tests (`_find_combine_mode` direct-call, end-to-end CLI routing, the DenseUnavailableError retry path). 24/24 pass.
- Full downstream suite re-run green, **128/128**: `test_reranker` (18), `test_reranker_hybrid` (14), `test_eval_late_rerank_quality` (20), `test_search_semantic_rerank` (5), `test_semantic_search_flag` (18, incl. real dense model), `test_bm25_search_flag` (3), plus the three files above.
- `ruff check` / `ruff format --check --preview` / `mypy` on all changed files: clean.

## Scope note

Ran targeted Python tests only (CPU-safe, shared box) — no cargo/rustc/maturin, no e2e Rust-compiling suites. Full CI matrix will run on this PR.

Leaving this **DRAFT** — awaiting the coordinator's re-verification of the fold's numbers before un-draft.

### DIFF
--- a/src/tensor_grep/cli/main.py
+++ b/src/tensor_grep/cli/main.py
@@ -14,7 +14,7 @@ from datetime import UTC, datetime
 from functools import lru_cache
 from pathlib import Path
 from tempfile import TemporaryDirectory
-from typing import TYPE_CHECKING, Any, cast
+from typing import TYPE_CHECKING, Any, Literal, cast
 from uuid import uuid4
 
 # Rich's legacy Windows renderer can raise EINVAL when long help is piped through
@@ -4261,9 +4261,45 @@ def _find_dense_weight(query: str) -> float:
     # genuinely multi-word (space-separated) query is NL and gets the resolved `weight` above.
     # Whitespace, not `split_terms` morphemes -- see the module comment above `_FIND_DENSE_WEIGHT_ENV`
     # for the dogfood finding (#191) this fixes.
-    if len(query.split()) <= 1:
+    if _find_is_single_token_query(query):
         return _FIND_DENSE_WEIGHT_DEFAULT
     return weight
+
+
+def _find_is_single_token_query(query: str) -> bool:
+    """Shared literal/identifier-query predicate for `tg find`'s two query-adaptive knobs --
+    `_find_dense_weight`'s dense_weight (above) AND `_find_combine_mode`'s RRF combine mode
+    (below). A single whitespace-free token (or an empty/whitespace-only query, whose ``split()``
+    -> ``[]``) is a literal identifier lookup; anything with 2+ whitespace-separated words is NL.
+    Whitespace, not `split_terms` morphemes -- see the module comment above
+    `_FIND_DENSE_WEIGHT_ENV` for the #191 dogfood finding this rule fixes. Factored out to a single
+    canonical definition so the two knobs can never silently disagree on what counts as literal.
+    """
+    return len(query.split()) <= 1
+
+
+def _find_combine_mode(query: str) -> Literal["sum", "max"]:
+    """Query-adaptive RRF `combine` mode for `tg find`'s `rank_chunks` calls (accuracy-leg
+    max-fusion regression fix, Opus-gate finding on PR #717): `combine="max"` (`rank_chunks`'s own
+    default, see its docstring) lifts genuinely multi-word/NL queries (ndcg@10 +62.6% on the frozen
+    40-query golden set) but REGRESSES single-token literal/identifier lookups -- measured on
+    `benchmarks/datasets/literal_golden.jsonl` (10 queries, `dense_weight=1.0` as
+    `_find_dense_weight` already protects them at): sum scores a perfect 1.0 ndcg@10 (as it always
+    has); max drops to 0.9631, a real -0.0369 regression. Mechanism: a literal query's true answer
+    is frequently ranked #1 by BOTH the bm25 leg and the dense leg independently -- `"sum"`'s
+    per-leg-agreement bonus is exactly the confirming signal `"max"` discards, letting a
+    single-leg-only competitor tie the true answer's now-undiscounted best-single-term score.
+
+    Routes on the EXACT SAME whitespace predicate `_find_dense_weight` already uses
+    (`_find_is_single_token_query`) so the two adaptive knobs can never disagree on what counts as
+    literal: a single whitespace-free token -> `"sum"` (byte-identical to the pre-max-flip
+    behavior, recovering the regression); a genuinely multi-word query -> `"max"` (keeps the NL
+    win, including composed with the adaptive `dense_weight=5.0`, which reaches the dense-alone
+    ndcg@10 ceiling exactly). This is scoped to `tg find` only -- `rerank_hybrid` (the `tg search
+    --semantic` path) never passes `combine`, so it is unaffected and stays on the plain "max"
+    default.
+    """
+    return "sum" if _find_is_single_token_query(query) else "max"
 
 
 def _find_representative_line(chunk: "Chunk", query_terms: set[str]) -> tuple[int, str]:
@@ -4466,6 +4502,7 @@ def _execute_find(
             dense_index=dense_index,
             late_reranker=late_reranker,
             dense_weight=_find_dense_weight(query),
+            combine=_find_combine_mode(query),
         )
     except DenseUnavailableError as exc:
         # F1 (Opus-gate blocker; mirrors `_apply_semantic_rerank`'s own query-time catch,
@@ -4491,6 +4528,7 @@ def _execute_find(
             dense_index=None,
             late_reranker=None,
             dense_weight=_find_dense_weight(query),
+            combine=_find_combine_mode(query),
         )
     if late_fallback_reason:
         result.rank_fallback_reason = (
--- a/src/tensor_grep/core/reranker.py
+++ b/src/tensor_grep/core/reranker.py
@@ -15,6 +15,7 @@ import os
 import sys
 import threading
 from collections import defaultdict
+from typing import Literal
 
 from tensor_grep.core.result import SearchResult
 from tensor_grep.core.retrieval_bm25 import Bm25Index
@@ -223,6 +224,7 @@ def rank_chunks(
     late_reranker: LateReranker | None,
     k: int = DEFAULT_K,
     dense_weight: float = 1.0,
+    combine: Literal["sum", "max"] = "max",
 ) -> tuple[list[int], str | None]:
     """Fuse BM25 [+ dense] [+ path] chunk rankings via RRF, then optionally MaxSim-rerank the head
     via ``late_reranker`` -- the pure, fail-closed rank core shared by :func:`rerank_hybrid` and
@@ -243,6 +245,20 @@ def rank_chunks(
     ``dense_index is None``: there is no dense leg to weight. Callers own picking a non-default
     value; `tg find` is the only current caller that ever does (``cli/main.py``'s
     ``_find_dense_weight``, itself gated behind ``TG_FIND_DENSE_WEIGHT`` and default-OFF).
+
+    ``combine`` (accuracy-leg campaign; passed straight through to
+    :func:`~tensor_grep.core.retrieval_fusion.reciprocal_rank_fusion`, see its own docstring for
+    the full derivation): ``"max"`` (this function's default, matching that function's own default
+    -- an omitted kwarg here is therefore byte-identical to one there) lifts NL/vocabulary-mismatched
+    queries (ndcg@10 +62.6% on the frozen 40-query golden set) but REGRESSES single-token
+    literal/identifier lookups (measured on ``benchmarks/datasets/literal_golden.jsonl``: sum=1.0
+    exact vs max=0.9631) -- a literal query's true answer is often independently ranked #1 by BOTH
+    legs, and ``"sum"``'s per-leg-agreement bonus is exactly the signal ``"max"`` discards. `tg
+    find` (``cli/main.py``'s ``_find_combine_mode``) is the one caller that ever passes a
+    non-default value, choosing per-query on the SAME whitespace predicate
+    ``_find_dense_weight`` already uses for its own knob, so the two adaptive decisions can never
+    disagree on what counts as literal. :func:`rerank_hybrid` (the ``tg search --semantic`` path)
+    never passes this kwarg, so it always gets the default ``"max"`` -- unaffected by this fix.
 
     Returns ``(fused_order, late_fallback_reason)``:
 
@@ -272,7 +288,7 @@ def rank_chunks(
             rankings.append(path_ranking)
             weights.append(PATH_CHANNEL_WEIGHT)
 
-    fused_order = reciprocal_rank_fusion(rankings, k=k, weights=weights)
+    fused_order = reciprocal_rank_fusion(rankings, k=k, weights=weights, combine=combine)
 
     # T5/T6: the late-interaction splice. Order-only over `fused_order`'s chunk indices -- same
     # matches, same membership, same JSON shape (design doc "The seam"). `late_reranker=None`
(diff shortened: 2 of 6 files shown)


oimiragieo:
**Independent Opus gate: SHIP-WITH-NITS → MUST-FIX + regression-fix folded → ready.**

The independent Opus gate (framed to falsify) verified all correctness targets on this load-bearing shared-ranking default flip: rrf arm ndcg@10 **0.3047→0.4953 (+62.6%) reproduced live at both endpoints** with dense/bm25 byte-unchanged, `combine="sum"` byte-reproduces the frozen baseline, **capsule 16/16 HARD floor held**, `--rank`/`--bm25` structurally unaffected (`rerank_by_bm25` has no dense leg), and the test teeth are **mutation-proven** both directions.

It also **falsified the build's "never worse than sum" overclaim**: on the committed `literal_golden.jsonl` slice, the max default scored 0.9631 vs sum 1.0 (−0.0369) — the 6-query dogfood + 40q golden set were NL-only and never covered the literal shape. Both folds landed (commit `f52fe50`):

1. **MUST-FIX (PR body):** overclaim corrected — the literal regression, the identifier wash, and the NL-only coverage are now stated plainly.
2. **Regression eliminated (not just documented):** single-whitespace-free-token queries now route to `combine="sum"` in `rank_chunks` via `_find_combine_mode` (reusing `_find_dense_weight`'s existing whitespace classifier), so literal lookups keep the sum optimum while multi-word NL keeps the max win. `reciprocal_rank_fusion`'s default stays `"max"` (verified — retrieval_fusion.py:39), so the scorecard's rrf arm is preserved.

**Post-fold numbers (build-agent verified, coordinator-confirmed on the crux):** rrf arm still **0.4953** (default untouched); `literal_golden` recovered to **1.0** via the route; NL find still hits the dense ceiling 0.6026; **capsule 16/16 unchanged**; 5 new route tests + 128/128 downstream green; ruff/format/mypy clean. (Honest note carried from the build: the literal regression is partly a harness-chunk-index-convention artifact — the real `_execute_find` walk gave 1.0 for both modes — so the route is a conservative safety-add.)

Coordinator verification: default-max preservation (scorecard crux) confirmed by direct read; route classifier correct; CI 0-fail. Un-drafting; enters the drain once CI is fully green.

PR #745, opened 2026-07-25
fix(search): honor root .gitignore in both Python rg-passthrough paths outside git repos (#269)

oimiragieo:
## Summary

Task #269: the third implementation of the #264 defect. Real `rg`'s own `.gitignore`
auto-discovery requires `require_git=true` by default, so a root `.gitignore` is silently a
no-op outside a git repository. PR #744 fixed this for the **compiled native** `tg` binary's
rg-passthrough (`rust_core/src/rg_passthrough.rs`), but explicitly left the **Python-only**
channel (pip/uvx installs with no native binary) unfixed and out of scope -- this PR closes
that gap.

## Step 1 -- the map (every Python rg-forwarding path)

Confirmed by direct tracing (monkeypatched both candidates and ran real invocations), not
assumed:

| # | `file:line` | Reached when | Pre-fix defect |
|---|---|---|---|
| 1 | `src/tensor_grep/cli/bootstrap.py:1088` (`_run_rg_passthrough`, via the forward-branch at `bootstrap.py:1256-1260`) | A bare plain-text `tg search PATTERN [.]` with no complicating flags. Confirmed this fires regardless of native-binary presence -- see the mechanism note below. | Yes -- raw argv forwarded to real `rg` verbatim, zero `--ignore-file` injection. |
| 2 | `src/tensor_grep/backends/ripgrep_backend.py` (`RipgrepBackend._build_cmd`, shared by `.search()`, `.search_passthrough()`, `._search_counts()`, `._search_files_with_matches()`) | Any search that bootstrap punts to `cli/main.py`'s full CLI (e.g. bare `--json`, since `--json` is in bootstrap's `_TG_ONLY_SEARCH_FLAGS`) and lands on `RipgrepBackend`, i.e. whenever no native binary is discoverable. | Yes -- same missing injection; only the pre-existing *explicit* `config.ignore_file` (`--ignore-file` CLI flag) was forwarded. |

Both were empirically confirmed broken with the same live repro:
non-git dir, root `.gitignore` excluding `skipme.txt`, both files containing `needle`,
`TG_DISABLE_NATIVE_TG=1`:
- `python -m tensor_grep search needle .` -> both files (bug)
- `python -m tensor_grep search --json needle .` -> both files (bug, `routing_backend: "RipgrepBackend"`)

**No third Python-only path exists.** `_build_cmd` is shared by every `RipgrepBackend` entry
point, so fixing it once covers `.search()`, `.search_passthrough()`, `._search_counts()`, and
`._search_files_with_matches()` uniformly.

### The mechanism behind the whole bug family (independently confirmed, matches sibling #744/c597b85)

`_can_delegate_to_native_tg_search` (`bootstrap.py:456-464`) only forwards a search to the
compiled native binary when argv contains one of `{--cpu, --force-cpu, --json, --ndjson,
--gpu-device-ids}`, or `TG_RUST_FIRST_SEARCH=1` (default off, `bootstrap.py:292-294`). A bare
plain-text search has none of those triggers, so bootstrap **always** falls through to
`_run_rg_passthrough` -- regardless of whether a native binary is discoverable. Proven with a
native binary present (not just absent) in
`tests/unit/test_cli_bootstrap.py::test_main_entry_bare_plain_text_search_bypasses_native_delegation_even_when_native_binary_present`.

`--json` **is** a supported trigger, so with a native binary present it reaches the native
engine directly (already correct per #127, not via rg-passthrough at all). That's the real
generator of "an output-format flag changes the file set": the flag doesn't touch
ignore-handling, it changes **which engine runs**. Flagging this for visibility per the
coordinator's request -- **not fixing it here**; whether a delegation gate should be keyed on
output-format flags at all is a separate design question.

## Step 2 -- the fix

New shared helper `src/tensor_grep/cli/rg_root_ignore.py::root_ignore_file_args()`, ported in
shape from `rust_core/src/rg_passthrough.rs::root_ignore_file_args` (branch
`fix/rg-passthrough-root-ignore-264`) rather than reinvented, so all three implementations
(Rust, and these two Python call sites) cannot silently diverge:
- Root-only scope (mirrors the native engine's `add_ignore` trio) -- no parent/nested walk.
- `.ignore` / `.gitignore` / `.rgignore`, ascending precedence order.
- `no_ignore` / `no_ignore_files` suppress all three (rg cancels `--ignore-file` outright on
  `--no-ignore-files`, so this short-circuits before any filesystem check).
- `no_ignore_vcs` skips only `.gitignore`; `no_ignore_dot` skips only `.ignore`/`.rgignore`
  (rg's own documented per-flag scope, reused from #744's live verification against rg 15.1.0,
  not re-derived).
- Existence-checked (`Path.is_file()`) before emitting `--ignore-file`, so rg never errors on a
  missing file and a pre-enumerated file list (the rare `--files-without-match` case) safely
  no-ops instead of emitting a wrong flag.

Wired into both call sites:
- `bootstrap.py::_run_rg_passthrough` -- roots come from `_search_path_args(search_args)` (the
  module's own existing raw-argv positional-path extractor, already used by
  `_search_args_paths_defaulted`), flags detected via simple `in search_args` membership
  (matching the existing style throughout this module). The emitted `--ignore-file <path>`
  operands are **prepended**, not appended, so they land before any `--` end-of-options
  sentinel the user's own argv might contain.
- `ripgrep_backend.py::RipgrepBackend._build_cmd` -- roots come from `file_path` (the same value
  real rg is about to walk), flags from the existing `SearchConfig` fields.

**Argv safety**: `--ignore-file <path>` is emitted as two adjacent argv tokens (never `=`-joined
or shell-interpolated), the same convention the pre-existing explicit `config.ignore_file`
handling in `ripgrep_backend.py` and the Rust reference already use -- a flag-value pair is
unambiguous to rg's parser regardless of what the value starts with, unlike a bare trailing
positional (which is what the codebase's `--` sentinel convention specifically protects
against elsewhere, e.g. `RipgrepBackend._append_search_paths`).

**No behavior change inside a git repo** -- verified live: real rg already auto-discovers the
same root `.gitignore` there, so the addition is redundant-not-harmful (ranks below rg's own
auto-discovered rules).

## Step 3 -- bidirectional tests

Unit (`tests/unit/test_rg_root_ignore.py`, `test_cli_bootstrap.py`, `test_ripgrep_backend.py`)
and e2e (`tests/e2e/test_python_rg_passthrough_root_ignore_outcome.py`) coverage mirroring
#744's Rust-side test matrix (injection, empty-when-absent, `no_ignore`/`no_ignore_files`/
`no_ignore_vcs`/`no_ignore_dot` gating, multi-root coverage).

**Measured pre-fix baseline** (`git stash` on just the two source fixes, tests kept): of the
new discriminating assertions, **8 unit-test failures** (4 in each of `test_cli_bootstrap.py`
and `test_ripgrep_backend.py` -- every assertion checking `--ignore-file` is *present*) and
**3 e2e failures** (`test_plain_text_search_excludes_gitignored_file_via_python_bootstrap_passthrough`,
`test_json_search_excludes_gitignored_file_via_python_ripgrep_backend`,
`test_plain_text_and_json_agree_on_the_same_one_file_set`) -- all confirmed to return both files
including `skipme.txt`. The negative-control tests (`--no-ignore*` escape hatches, the
no-ignore-files-present empty case, the git-repo cell) correctly passed in **both** arms, as
expected -- they pin pre-existing, unrelated-to-this-fix behavior, not this fix's discriminating
claim. All tests pass post-fix; fix restored and re-verified.

## Skip-removal status

`tests/e2e/test_search_root_ignore_file_outcome.py` (the file whose native-binary skip
documents this exact gap) lives on **PR #744's branch**, not yet merged to `main` as of this
PR -- it does not exist on this branch. Per the task brief, rather than editing across branches,
this PR adds a **parallel Python-path case**:
`tests/e2e/test_python_rg_passthrough_root_ignore_outcome.py`, which forces the no-native-binary
condition deterministically (`TG_DISABLE_NATIVE_TG=1`) rather than skipping -- it never needs to
skip on that account. **Dependency note for reviewers**: once #744 merges, its outcome test's
docstring (which references this exact gap by name) should be revisited to point at this PR's
closure; no direct file conflict expected since the two test files are independent.

## Coordination

- Stayed entirely inside `bootstrap.py` (search-passthrough function only) and
  `ripgrep_backend.py`; did not touch `_force_utf8_streams` (#743's area) or `rust_core/`
  (#742/#744's area).
- New file `src/tensor_grep/cli/rg_root_ignore.py` has no other owners in flight.

Generated with Claude Code

### DIFF
--- a/.claude/rg_argv_differential_fuzz.py
+++ b/.claude/rg_argv_differential_fuzz.py
@@ -1,0 +1,258 @@
+#!/usr/bin/env python3
+"""Differential fuzz: `bootstrap._search_path_args_raw` vs an independent model of rg's grammar.
+
+WHY THIS EXISTS
+---------------
+Task #269 (root-`.gitignore` honoring on the Python-only rg-passthrough paths) went through FOUR
+rounds in which every round found an argv form nobody had enumerated by hand:
+
+  1. `-u`/`-uu`/`--unrestricted` ungated  -> `-u` became STRICTER than no flag
+  2. `-f`/`--file`/`-e<attached>` not treated as pattern sources -> wrong root's ignore file
+  3. `-ieneedle` (mid-bundle attached)    -> same, introduced by round 2's own fix
+  4. `PATH -eneedle` (positional BEFORE the flag) -> same, introduced by round 3's own fix
+
+Hand-written enumerations kept missing a dimension. This harness replaces the enumeration with a
+whole-grammar model: it parses argv the way rg 15.1.0 documents itself to (via `rg --help`'s own
+flag tables), computes the PATH positionals rg would search, and compares that against what
+`_search_path_args_raw` computes. Every disagreement is a candidate wrong-root -> wrong
+`--ignore-file` injection -> silently wrong file set.
+
+The reference model here is deliberately the ONLY grammar model in the repo. Do not write a
+second one: two independently-written models disagree with each other and neither is
+authoritative. If rg's grammar changes, update `SHORT_VALUE` / `LONG_VALUE` below from
+`rg --help` and re-run.
+
+USAGE
+-----
+    python .claude/rg_argv_differential_fuzz.py
+    python .claude/rg_argv_differential_fuzz.py --seed 20260725 --iterations 60000
+    python .claude/rg_argv_differential_fuzz.py --src path/to/src   # test another checkout
+
+Exit code 0 = zero disagreements (the closing gate). Exit code 1 = disagreements, printed with
+a minimal reproducer per distinct shape so a failure is actionable, not just a count.
+
+The run is deterministic for a given `--seed`, so a reported failure reproduces exactly.
+
+REFRESHING THE FLAG TABLES (do this when bumping the pinned rg)
+--------------------------------------------------------------
+    rg --help | grep -oE "^\\s+-[a-zA-Z] [A-Z<]"          # -> SHORT_VALUE
+    rg --help | grep -oE "^\\s+(-[a-zA-Z], )?--[a-z0-9-]+=" # -> LONG_VALUE
+"""
+
+from __future__ import annotations
+
+import argparse
+import importlib.util
+import itertools
+import random
+import sys
+from pathlib import Path
+
+# --------------------------------------------------------------------------------------
+# rg 15.1.0 grammar tables, taken from `rg --help`'s own output (see module docstring).
+# --------------------------------------------------------------------------------------
+
+#: Short flags that take a value. `rg --help` prints these as `-X ARG, --long=ARG`.
+#: A value-taking short flag inside a cluster SWALLOWS the remainder of the token as its value
+#: (`-ite` == `-i -t e`, verified live: `rg -ite needle .` -> "unrecognized file type: e"), or,
+#: when it is the token's LAST character, consumes the NEXT argv token (`-ie needle`).
+SHORT_VALUE = set("ABCEMdefgjmrtT")
+
+#: Long flags that take a value. `rg --help` prints these with an `=ARG` suffix.
+LONG_VALUE = {
+    "--after-context",
+    "--before-context",
+    "--color",
+    "--colors",
+    "--context",
+    "--context-separator",
+    "--dfa-size-limit",
+    "--encoding",
+    "--engine",
+    "--field-context-separator",
+    "--field-match-separator",
+    "--file",
+    "--generate",
+    "--glob",
+    "--hostname-bin",
+    "--hyperlink-format",
+    "--iglob",
+    "--ignore-file",
+    "--max-columns",
+    "--max-count",
+    "--max-depth",
+    "--max-filesize",
+    "--path-separator",
+    "--pre",
+    "--pre-glob",
+    "--regex-size-limit",
+    "--regexp",
+    "--replace",
+    "--sort",
+    "--sortr",
+    "--threads",
+    "--type",
+    "--type-add",
+    "--type-clear",
+    "--type-not",
+}
+
+#: Flags that SUPPLY THE PATTERN. If any of these appears in the option region, rg does NOT
+#: treat the first bare positional as the pattern -- every bare positional is a PATH. rg's
+#: grammar is ORDER-INDEPENDENT here: `rg sub -eneedle` == `rg -eneedle sub` (verified live).
+PATTERN_SOURCE_SHORT = set("ef")
+PATTERN_SOURCE_LONG = {"--regexp", "--file"}
+
+
+def rg_roots(argv: list[str]) -> list[str]:
+    """Return the PATH positionals rg would search. Empty list == implicit cwd root."""
+    positionals: list[str] = []
+    pattern_from_flag = False
+    end_of_opts = False
+    i = 0
+    while i < len(argv):
+        arg = argv[i]
+        if end_of_opts:
+            positionals.append(arg)
+            i += 1
+            continue
+        if arg == "--":
+            end_of_opts = True
+            i += 1
+            continue
+        if arg.startswith("--"):
+            name, sep, _value = arg.partition("=")
+            if name in PATTERN_SOURCE_LONG:
+                pattern_from_flag = True
+            if not sep and name in LONG_VALUE:
+                i += 1  # value lives in the next argv token
+            i += 1
+            continue
+        if arg.startswith("-") and len(arg) > 1:
+            cluster = arg[1:]
+            for offset, char in enumerate(cluster):
+                if char in SHORT_VALUE:
+                    if char in PATTERN_SOURCE_SHORT:
+                        pattern_from_flag = True
+                    if offset == len(cluster) - 1:
+                        i += 1  # value lives in the next argv token
+                    break  # the remainder of the token is this flag's value
+            i += 1
+            continue
+        positionals.append(arg)
+        i += 1
+    if pattern_from_flag:
+        return positionals
+    return positionals[1:]  # the first bare positional is the PATTERN
+
+
+# --------------------------------------------------------------------------------------
+# Corpus
+# --------------------------------------------------------------------------------------
+
+BOOL_SHORTS = list("inFvswxaqlcHINopzSPULb")
+VALUES = ["5", "needle", "pats.txt", "*.py", "py", "utf-8", "-weird"]
+PATTERNS = ["needle", "unwrap", "u", "g*.py"]
+PATHS = [".", "otherdir", "sub", "--"]
+CLUSTER_PREFIXES = ["", "i", "n", "iv", "vF", "zi", "s", "in", "wxa"]
+
+
+def build_corpus(seed: int, iterations: int) -> list[list[str]]:
+    """Systematic sweep (every prefix x value-flag x attach/separate x tail) + seeded fuzz."""
+    cases: list[list[str]] = []
+    for prefix, flag in itertools.product(CLUSTER_PREFIXES, sorted(SHORT_VALUE)):
+        for value in ("needle", "5", "*.py", "py", "utf-8"):
+            cases.append([f"-{prefix}{flag}{value}", "otherdir"])
+            cases.append([f"-{prefix}{flag}", value, "otherdir"])
+            cases.append([f"-{prefix}{flag}{value}"])
+            cases.append([f"-{prefix}{flag}", value])
+            cases.append([f"-{prefix}{flag}", value, "needle", "otherdir"])
+            cases.append([f"-{prefix}{flag}", value, "needle", "--", "otherdir"])
+            # ORDERING dimension (round 4): the PATH ahead of the pattern-source flag.
+            cases.append(["otherdir", f"-{prefix}{flag}{value}"])
+            cases.append(["otherdir", f"-{prefix}{flag}", value])
+
+    rnd = random.Random(seed)
+    alphabet = (
+        [f"-{c}" for c in sorted(SHORT_VALUE) + BOOL_SHORTS]
+        + [f"-{a}{b}" for a in "invFszwx" for b in sorted(SHORT_VALUE) + BOOL_SHORTS]
+        + [f"-{a}{v}" for a in sorted(SHORT_VALUE) for v in VALUES]
+        + [f"-{p}{a}{v}" for p in "inz" for a in sorted(SHORT_VALUE) for v in ("needle", "5", "py")]
+        + sorted(LONG_VALUE)
+        + [f"{k}={v}" for k in sorted(LONG_VALUE) for v in ("needle", "py", "5")]
+        + ["--no-ignore", "--unrestricted", "--files", "--json", "--", "--hidden"]
+        + PATTERNS
+        + PATHS
+        + VALUES
+    )
+    for _ in range(iterations):
+        cases.append([rnd.choice(alphabet) for _ in range(rnd.randint(1, 6))])
+    return cases
+
+
+# --------------------------------------------------------------------------------------
+
+
+def load_bootstrap(src: Path):
+    sys.path.insert(0, str(src))
+    spec = importlib.util.spec_from_file_location(
+        "_fuzz_bootstrap", src / "tensor_grep" / "cli" / "bootstrap.py"
+    )
+    if spec is None or spec.loader is None:  # pragma: no cover - defensive
+        raise SystemExit(f"cannot import bootstrap.py from {src}")
+    module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(module)
+    return module
+
+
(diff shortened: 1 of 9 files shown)


oimiragieo:
## Independent-gate BLOCKING finding addressed (commit 68ed1e8)

Fixed exactly as prescribed:

1. `rg_root_ignore.py` — added `unrestricted: int = 0`, gated alongside `no_ignore`/`no_ignore_files`.
2. `ripgrep_backend.py` — forwards `config.unrestricted` into the new parameter.
3. `bootstrap.py` — extracted the raw-argv `-u` predicate into a named `_search_args_request_unrestricted` helper (reusing the existing `_SEARCH_ATTACHED_VALUE_SHORT_FLAGS` cluster-walk machinery from `_requires_full_cli`), shared by both `_search_args_request_unrestricted_generated_scan` and `_run_rg_passthrough` so they cannot drift. Clustered forms (`-iu`, `-nu`) are handled and unit-tested; `-tu` (attached-value flag swallowing `u` as its own value) is unit-tested as a non-match.
4. e2e parametrize widened to include `-u` on both plain-text and `--json`; added the non-blocking `--json`/`-u`+`--no-ignore` outcome control for `RipgrepBackend` that was previously argv-only coverage.

**Bidirectional proof**: stashed only this follow-up's source changes, confirmed 12 unit + 2 real-subprocess e2e assertions fail against the prior committed state (all reproducing the exact reported regression — `-u needle .` returning only the non-ignored file instead of all files). Restored; all 181 tests across the 4 affected files pass. `ruff format --preview` + `ruff check` clean.

**Two flagged, not-fixed items, now resolved by reading (no build needed):**
- *Native-binary-present `-u` cell*: `rust_core/src/main.rs`'s `SEARCH_PYTHON_PASSTHROUGH_FLAGS` (consumed at `main.rs:1632`, `:1765`, and others) lists both `-u` and `--unrestricted` — the compiled native binary unconditionally forwards any `-u` search to this Python sidecar rather than handling it in its own rg-passthrough. So this fix *does* cover the native-present channel end-to-end; no separate Rust-side verification is needed for `-u` specifically.
- *Does #744's Rust `root_ignore_file_args`/`_build_ripgrep_command` need `-u` handling?* No — confirmed `RipgrepSearchArgs` (the struct that Rust function operates on) has no `unrestricted` field at all, because `-u`/`--unrestricted` is structurally routed to Python before Rust's `execute_ripgrep_search` ever runs. Worth a note on #744 for the record, but no action needed there.

Also added the low-priority comment on the unhandled `--ignore`/`--ignore-vcs` re-enable-flag gap per the non-blocking note (documented as an intentional under-emit, safe direction, not gated).

oimiragieo:
## Re-gate BLOCKING findings addressed (commit 61b93fc)

**FIX-1 (silent, `--file`/`-f`/`-e<attached>` misrouting root-ignore injection)** — added `_SEARCH_PATTERN_SOURCE_FLAGS` / `_SEARCH_PATTERN_SOURCE_ATTACHED_SHORT_FLAGS`, checked in `_search_path_args_raw` before the generic value/attached-value branches, so `-f`/`--file`/`--file=`/attached `-fVAL`/`-eVAL` all correctly mark "a pattern is already supplied" instead of letting the real target PATH get misread as the bare pattern. Kept separate from the shared `_SEARCH_PATTERN_FLAGS` (used by `_regex_patterns_from_search_args` to extract literal pattern *text* — folding `-f`/`--file` in there would feed a filename into regex validation).

Re-ran the exact measured table:

| argv | before this commit | after this commit |
|---|---|---|
| `--file pats.txt otherdir` | a.txt only ❌ | a.txt, b.log ✅ |
| `-f pats.txt otherdir` | a.txt only ❌ | a.txt, b.log ✅ |
| `-fpats.txt otherdir` | a.txt only ❌ | a.txt, b.log ✅ |
| `-eneedle otherdir` | a.txt only ❌ | a.txt, b.log ✅ |
| `-e needle otherdir` (control) | both ✅ | both ✅ |

**FIX-2 (`-e` missing from `_SEARCH_ATTACHED_VALUE_SHORT_FLAGS`, introduced by the prior round's own extraction)** — added `-e`. Re-ran:

```
tg search --no-heading -eunicorn .
  before this commit : rc=2  Error: No such option '--no-heading'.
  after this commit  : rc=0  src/y.js, node_modules/pkg/x.js
tg search --no-heading -enicorn .   (control, no 'u')  -> rc=0 both before and after
```

Also added the parity checks requested: `-eunwrap`/`-eunicorn` → False; `-iu`/`-nu`/`-au` still → True; `-eg*.py` no longer misroutes `_requires_full_cli`.

**Bidirectional proof**: stashed only this round's source changes, confirmed 7 unit + 4 real-subprocess e2e assertions fail against the prior commit, all reproducing the measured tables above verbatim. Restored; 194 tests across the 4 affected files pass. `ruff format --preview` + `ruff check` clean.

**Non-blocking corrections applied**: `rg_root_ignore.py`'s docstring no longer claims a caller "cannot distinguish" the `--ignore*` re-enable flags (factually corrected — `SearchConfig` already carries and forwards `ignore`/`ignore_dot`/`ignore_files`/`ignore_vcs` elsewhere in `_build_cmd`; the gap is that neither call site threads that signal into *this* gate, a scope decision not an inability). Added a comment on `_build_cmd`'s `if config:` gate noting `search_passthrough(config=None)` forwards zero `SearchConfig`-derived flags (pre-existing, safe, never hit by either real call site — not unique to this injection).

oimiragieo:
## Delta-gate BLOCKING-3 + NB-1 + NB-2 addressed (commit b913f4e)

**BLOCKING-3** (mid-bundle `-ieVAL`/`-ie VAL` not caught by the FIX-1 round's `startswith` check) — replaced the narrow first-character check with the general bundled/clustered cluster walk `_requires_full_cli` and `_search_args_request_unrestricted` already use elsewhere in this file: scan past leading boolean short flags until the first attached-value short flag, consuming the remainder of the token (or the next argv token if it's the token's last character) as its own value. `-e`/`-f` mark a pattern as supplied; every other attached-value flag gets identical value-consumption handling, which **closes NB-1 in the same pass** (`-im 5 needle otherdir` no longer misreads "needle" as a path).

Pinned **both** the attached form and the "by-accident" separate-value form explicitly, per the gate's instruction — `-ie needle otherdir` produced the right final answer before this fix (the stray value token got swept up as the bare pattern instead of a path, which happened to leave the real PATH intact), so a test asserting only the final answer wouldn't discriminate a regression confined to that shape. Bidirectional check confirms this: 4 unit + 2 e2e assertions fail on the prior commit (the attached forms), while the by-accident separate-value forms and the NB-1 outcome test correctly pass in *both* arms (documented as pinning pre-existing/accidental behavior, not this fix's claim).

The now-fully-redundant `_SEARCH_PATTERN_SOURCE_ATTACHED_SHORT_FLAGS` tuple is removed (its one call site is replaced by the walk).

**NB-2** (stale line-number citations) — audited every `file.py:NNN` citation this task's commits introduced across `bootstrap.py`, `ripgrep_backend.py`, `rg_root_ignore.py`, and the three test files. Found and fixed 5 more instances beyond the one you flagged (all had drifted from this task's own later edits shifting lines below them — `_can_delegate_to_native_tg_search` cited as `:456-464` is actually at 487, `_prefer_rust_first_search` cited as `:292-294` is actually at 323, `_run_rg_passthrough` cited as `:1088` is actually at 1179, plus two internal cross-references within `bootstrap.py` itself). Rather than re-numbering (which would just drift again next edit), replaced every one with a name-only reference (function/field name, no line number) and left a one-line note at each site explaining why, so a future editor doesn't reintroduce the pattern.

**Verified**: 205 tests across the 4 affected files pass, `ruff format --preview` + `ruff check` clean, bidirectional proof re-run and holds.

oimiragieo:
## Final gate BLOCKING-1 + NON-BLOCKING-2 addressed (commit 2be893f)

**BLOCKING-1** (ordering dimension: PATH before the pattern-source flag) — made `_search_path_args_raw` genuinely two-pass. New `_search_args_contains_pattern_source_flag` pre-pass scans all of `search_args` (honoring `--`, correctly consuming other flags' values so an `-e`-shaped token that's actually another flag's *value* is never miscounted) to determine whether a pattern-source flag exists anywhere before `--`, seeding `regexp_pattern_seen`'s starting value instead of letting the single left-to-right extraction walk discover it mid-stream. Both passes share a new `_attached_cluster_value_offset` helper for the cluster-scan logic rather than re-deriving it twice — directly addressing the "two passes computing the same thing differently" shape that caused this exact bug.

Verified against your measured repro:
```
tg search sub -eneedle          -> sub/c.txt only (sub's own .gitignore honored, d.log excluded)
tg search --json sub -eneedle   -> sub/c.txt only (agrees)
tg search otherdir -eneedle     -> otherdir/a.txt, otherdir/b.log (otherdir has no ignore file — nothing injected, cwd's .gitignore does not leak)
```

Added the six-spelling parametrized test you specified (`_search_path_args_raw(["otherdir", "-eneedle"]) == ["otherdir"]` and its `-e VAL`/`-f FILE`/`-f FILE`/`--regexp=`/`--file=` siblings), a mid-bundle × ordering composition test, an explicit `-e needle -- sub` control (confirmed unaffected — still correct, as it was before), a consumed-flag-value negative control (`--replace -eattached` must not be misdetected as a pattern-source flag), and two e2e outcome rows checking plain-text/`--json` agreement in both the "target has its own ignore file" and "target has none" cases.

Bidirectional check: 11 unit + 2 real-subprocess e2e assertions fail against the prior commit (several via `AttributeError` since the pre-pass function didn't exist yet, the rest reproducing the exact wrong-file-set and #264-signature regressions from your report). 219 tests pass post-fix.

**On the fuzz harness** — per your suggestion, I'd like to run it rather than re-derive an equivalent from scratch. Could you share the independent-parser + differential-fuzz harness (or point me at where it's checked in / how to invoke it)? I'll re-run it against `2be893f` and report the disagreement count.

**NON-BLOCKING-2 finished** — found 3 more stale citations your sweep's exact grep pattern missed (a different phrasing, `"lines ~NNN-NNN"` with no filename prefix, that my prior pattern-matching pass didn't catch): `bootstrap.py`'s `_search_args_request_unrestricted` docstring ("lines ~369-374 above"), `test_cli_bootstrap.py:691`'s "~line 369", and `test_ripgrep_backend.py:165`'s "line ~648" (confirmed actual site is `:667`, off by 19 as you found). All converted to name-only references, matching the other 14. Also strengthened `bootstrap.py`'s "the roots this computes always match what real rg will actually search" comment (the claim your report noted was falsified) to explicitly scope it to post-BLOCKING-1 and explain why it was briefly untrue.

**NON-BLOCKING-3** — noted for `docs/CONTRACTS.md`, no action taken this PR per your guidance (contract-consistent with the already-shipped Rust behavior).

All 219 tests pass, `ruff format --preview` + `ruff check` clean.

oimiragieo:
## Fuzz-round BLOCKING-2 addressed + fuzz harness wired into CI (commit 0025818)

**BLOCKING-2** (offset unified, consumption was not) — the pre-pass now also sets `skip_next = True` when the attached-value char is the token's last character, exactly mirroring what the extraction walk already did. Two-line fix at the exact site you cited. Added `-ir -e needle` / `-ig -fpats.txt sub` to the negative-control parametrization (now 3 rows: the working long-separated-form control, plus the two mid-bundle rows that actually exercise the broken arm) with a comment documenting why the single long-form control proved nothing about the short-bundle path.

The same fuzz run (after BLOCKING-2) surfaced one more pre-existing gap: `--generate` takes a value in real rg but was only listed in `_TG_ONLY_SEARCH_FLAGS` (routing), never `_SEARCH_FLAGS_WITH_VALUES` (value-consumption in the positional walk) — added it there too.

**Fuzz harness wired into CI** — copied `.claude/rg_argv_differential_fuzz.py` into this branch and added it as a new step in the `static-analysis` job (already a `release`-gating job) in `.github/workflows/ci.yml`, running with `--seed 20260725 --iterations 65000` (70,040 total cases including the systematic sweep). Caught one real bug before it landed: my initial `- name: rg argv grammar differential fuzz (task #269)` had an unquoted `#269` that YAML parses as a mid-line comment, silently truncating the step name — quoted the string to fix it, verified with a `yaml.safe_load` round-trip.

**Verified**: the harness reports **0/70040 disagreements** across three seeds (20260725, 1, 42) against the fixed code. Bidirectional check: both new negative-control rows fail against the prior commit, and the fuzz harness itself shows dozens of disagreements against that same prior state. 221 tests pass, ruff clean including the new harness file.

**Noted, no action needed**: the bidirectional-count correction (14 not 13, conservative direction) and the citation resolution (all 3 flagged instances gone, the 2 PR-introduced ones resolve correctly, the remaining 11 are pre-existing/out-of-scope) — thank you for tracking those down precisely.

PR #267, opened 2026-06-25
docs: de-brittle SESSION_HANDOFF/AGENTS/CONTINUATION release ledgers (follow-up to #264)

oimiragieo:
Extends the #264 governance de-brittling to the remaining handoff docs (the documented follow-up).

Removed the per-version release-proof ledger lines — the "Closed vX gap: \<PR/commit\>" changelog + "Latest verified release proof commit" lines — from `docs/SESSION_HANDOFF.md` (−22), `AGENTS.md` (−3), `docs/CONTINUATION_PLAN.md` (−3), and dropped the 11 per-version `vX.Y\` release …` pins from `test_handoff` (a hand-maintained changelog that drifted; `CHANGELOG.md` + git log are canonical).

**All** current-state, thematic ("Closed \<theme\> gap"), and behavioral/capability checks are retained. 230 doc-governance + release-asset + readiness tests pass; ruff clean. Docs only → **no release**.

PR #750, opened 2026-07-25
fix(search): thread --no-ignore-vcs into the native --json/--ndjson walker (#267)

oimiragieo:
## Summary

Closes #267. Confirmed the claim reproduces on the currently-shipped binary and fixed it.

`--no-ignore-vcs` (rg's "don't respect `.gitignore`, but keep `.ignore`/`.rgignore`" flag) is
silently dropped whenever a search is also routed through the native `--json`/`--ndjson` engine
(`NativeCpuBackend`, `routing_reason=json_output`). The identical flag on a plain-text search
(no `--json`) is honored correctly. So adding a pure output-format flag changes the FILE SET —
exactly the #270 root-generator pattern the #264/#269/#272 siblings already closed for their own
cells.

## Root cause

`NativeSearchConfig` (`rust_core/src/native_search.rs`) never had a `no_ignore_vcs` field at all.
`build_walk_builder` only ever branched on the blanket `no_ignore`:

```rust
if config.no_ignore {
    builder.ignore(false); ...
} else {
    for root in roots {
        for ignore_name in [".ignore", ".gitignore", ".rgignore"] {
            // .gitignore was ALWAYS added here, no_ignore_vcs was never consulted
            ...
        }
    }
}
```

`route_search` sends every `--json`/`--ndjson` search to exactly this engine
(`structured_output` arm → `RoutingDecision::native_cpu_json`), bypassing the rg-passthrough
engine (`rg_passthrough.rs::root_ignore_file_args`, fixed for this exact non-git-repo gap by
#264/#744) entirely. `search_prefers_ripgrep_passthrough` short-circuits to `false` the instant
`args.json` is set, before it ever reaches the `no_ignore_vcs` check further down — so the
`--no-ignore-vcs`-aware rg-passthrough path is architecturally unreachable once `--json` is on
the command line.

## How I verified (bidirectional, tg-vs-tg, one flag at a time)

Per the task's method: a non-git fixture dir (root `.gitignore` = `*.log`, `sub/.gitignore` =
`*.txt`, sentinel files under each), comparing **file sets**, not match counts, using the
**published v1.98.8** `tg-windows-amd64-cpu.exe` (`gh release download v1.98.8`) — same
`build_walk_builder` body as `origin/main` (`git diff v1.98.3 HEAD -- rust_core/src/native_search.rs`
shows zero changes to that function outside this PR), so this binary is a faithful pre-fix oracle
for `origin/main`'s current behavior. `TG_DISABLE_NATIVE_TG` wasn't needed since the divergence
*is* the native-vs-native-engine split (`--verbose` prints `routing_backend`/`routing_reason` to
prove which of tg's own two engines actually served each request).

| search_args | routing_backend (from `--verbose`) | files returned |
|---|---|---|
| *(bare)* | `RipgrepBackend` / `rg_passthrough` | `keep_root.txt`, `sub/ignored_sub.txt` |
| `--no-ignore-vcs` | `RipgrepBackend` / `rg_passthrough` | `ignored_root.log`, `keep_root.txt`, `sub/ignored_sub.txt` — **correctly re-included** |
| `--json` | `NativeCpuBackend` / `json_output` | `keep_root.txt`, `sub/ignored_sub.txt` |
| `--json --no-ignore-vcs` | `NativeCpuBackend` / `json_output` | `keep_root.txt`, `sub/ignored_sub.txt` — **identical to bare `--json`, bug** |
| `--json --no-ignore` (control) | `NativeCpuBackend` / `json_output` | `ignored_root.log`, `keep_root.txt`, `sub/ignored_sub.txt` — **correctly re-included, proves only `no_ignore_vcs` specifically was dropped, not the whole ignore-file mechanism** |

The invariant `files(A) == files(A + renderer-flag)` holds for every OTHER flag combination
tested (bare vs `--json`, `--no-ignore` vs `--json --no-ignore`) — it specifically breaks for
`--no-ignore-vcs` + `--json`, matching the filed claim precisely (not the more sweeping "native
side drops the flag" wording from the issue title — the flag is present and parsed on the native
side at every site cited in the task; the gap is narrowly in `NativeSearchConfig`/
`build_walk_builder` not threading it through for the `--json`/`--ndjson`/GPU-CPU-fallback
routes).

## Fix

- `NativeSearchConfig` gains `no_ignore_vcs: bool` (default `false`).
- `build_walk_builder` skips only `.gitignore` (not `.ignore`/`.rgignore`) when it's set —
  mirroring `rg_passthrough.rs::root_ignore_file_args`'s existing per-filename scope, matching
  rg's own documented restriction of `--no-ignore-vcs` to source-control ignore files.
- Threaded from `SearchArgs.no_ignore_vcs` in `native_search_config_for_command` (the `--json`
  route) and from `GpuSearchParams.no_ignore_vcs` (already present on that struct, just never
  copied over) in `native_search_config_for_gpu_params` (the GPU→CPU-fallback route — same shape
  of bug, one call site over, fixed for consistency since the data was already sitting right
  there).
- `native_search_config_for_positional` (the bare `tg PATTERN PATH` shorthand) is untouched: it
  hardcodes `no_ignore: true` unconditionally already, so `no_ignore_vcs` is moot there.

## Tests

4 new unit tests in `rust_core/src/native_search.rs`, all asserting the **file set** returned by
`collect_walked_files` (the same function `run_native_search` calls), not the mechanism:
1. default behavior unchanged (`.gitignore` still excludes a match with `no_ignore_vcs=false`)
2. `--no-ignore-vcs` re-includes the `.gitignore`-matched file (the bug this PR fixes)
3. `--no-ignore-vcs` does NOT touch `.ignore`-matched files (rg's documented scope)
4. `--no-ignore` still overrides everything regardless of `no_ignore_vcs` (non-regression)

**Honesty on the bidirectional-oracle requirement**: CPU-SAFE forbids `cargo build/test` on this
shared box, so I could not literally run `cargo test` to watch test #2 go red on the pre-fix code
and green after. What I *did* do: (a) the live-binary table above is a genuine pre-fix execution
of the exact same `build_walk_builder` logic (verified byte-identical via `git diff` against
v1.98.3, and structurally identical to `origin/main` before this PR), so the RED direction is
execution-verified, just not via `cargo test`; (b) I hand-traced that reverting the one-line
`if ignore_name == ".gitignore" && config.no_ignore_vcs { continue; }` guard makes test #2's
`ignored.log` never get added to the walker, causing the exact assertion failure described in
the test's own comment. GitHub CI (`cargo test`) is the compile+execute oracle for the unit tests
themselves.

## Explicitly out of scope

- `rust_core/src/index.rs::collect_file_entries` (the warm `.tg_index` build path) has the
  identical shape of gap (`no_ignore` only, no `no_ignore_vcs` field) — but it's only reachable
  when an explicit `--index` flag or an existing warm index cache is in play, which my
  reproduction didn't exercise and which `route_search` checks *before* the `--json` branch this
  PR fixes. Flagging as a related residual, not fixing here, to keep this change scoped to what
  was actually reproduced.
- Engine selection itself (which flags route to `RipgrepBackend` vs `NativeCpuBackend`) is
  unchanged — per the task, that's the CEO-gated native-front-door question (#240/#48-architectural).

## Test plan
- [x] `rustfmt --check rust_core/src/native_search.rs rust_core/src/main.rs` — clean
- [x] Live-binary before/after table above (v1.98.8 release binary, same walker logic as this PR's base)
- [ ] CI `cargo test` (release-blocking gate, not run locally per CPU-SAFE)
- [ ] Independent gate before un-drafting (per house policy)

### DIFF
--- a/rust_core/src/main.rs
+++ b/rust_core/src/main.rs
@@ -8073,6 +8073,12 @@ fn native_search_config_for_command(
         max_depth: args.max_depth,
         count: args.count,
         no_ignore: args.no_ignore,
+        // Task #267: without this, `--json`/`--ndjson` (this function's `structured_output`
+        // callers) silently dropped `--no-ignore-vcs` -- `build_walk_builder` had no field to
+        // read it from -- while the identical flag on the non-structured-output route (real
+        // `rg` via `command_ripgrep_args`/`root_ignore_file_args`) honored it correctly. An
+        // output-format flag must never change the file set.
+        no_ignore_vcs: args.no_ignore_vcs,
         json: args.json,
         ndjson: args.ndjson,
         verbose: args.verbose,
@@ -8116,6 +8122,12 @@ fn native_search_config_for_gpu_params(
         max_depth: params.max_depth,
         count: params.count,
         no_ignore: params.no_ignore,
+        // Task #267: same gap as `native_search_config_for_command` -- this is the
+        // explicit-`--gpu-device-ids`-fallback-to-CPU route, which already carries
+        // `params.no_ignore_vcs` (used by the GPU engine itself) but was never threading it
+        // into the CPU-fallback `NativeSearchConfig`, silently dropping `--no-ignore-vcs` on
+        // this route the same way.
+        no_ignore_vcs: params.no_ignore_vcs,
         json: params.json,
         ndjson: params.ndjson,
         verbose: params.verbose,
--- a/rust_core/src/native_search.rs
+++ b/rust_core/src/native_search.rs
@@ -191,6 +191,17 @@ pub struct NativeSearchConfig {
     pub count: bool,
     pub crlf: bool,
     pub no_ignore: bool,
+    /// Task #267: mirrors `RipgrepSearchArgs::no_ignore_vcs` / `rg_passthrough::
+    /// root_ignore_file_args`'s `no_ignore_vcs` gate (rg docs: "--no-ignore-vcs" restricts
+    /// disabling to source-control ignore files, i.e. skip only `.gitignore`, unlike the
+    /// blanket `no_ignore`). Before this field existed, `build_walk_builder` had no way to
+    /// honor `--no-ignore-vcs` at all -- it only ever branched on `no_ignore` -- so any
+    /// caller that routes through this engine (`--json`/`--ndjson`, the structured_output
+    /// arm of `route_search`) silently ignored `--no-ignore-vcs` and kept excluding
+    /// `.gitignore`-matched files, even though the identical flag combination on the
+    /// non-structured-output path (rg passthrough, `rg_passthrough.rs::root_ignore_file_args`)
+    /// correctly re-includes them. An output-format flag must never change the file set.
+    pub no_ignore_vcs: bool,
     pub line_number: bool,
     pub with_filename: bool,
     pub replace: Option<String>,
@@ -232,6 +243,7 @@ impl Default for NativeSearchConfig {
             count: false,
             crlf: false,
             no_ignore: false,
+            no_ignore_vcs: false,
             line_number: true,
             with_filename: false,
             replace: None,
@@ -1659,6 +1671,13 @@ fn build_walk_builder(config: &NativeSearchConfig, roots: &[PathBuf]) -> Result<
     } else {
         for root in roots {
             for ignore_name in [".ignore", ".gitignore", ".rgignore"] {
+                // Task #267: mirror `rg_passthrough::root_ignore_file_args`'s per-filename
+                // `no_ignore_vcs` scope -- rg's own docs restrict `--no-ignore-vcs` to
+                // source-control ignore files, so only `.gitignore` is skipped here; `.ignore`
+                // and `.rgignore` stay honored exactly like the rg-passthrough engine.
+                if ignore_name == ".gitignore" && config.no_ignore_vcs {
+                    continue;
+                }
                 let ignore_path = root.join(ignore_name);
                 if ignore_path.is_file() {
                     builder.add_ignore(ignore_path);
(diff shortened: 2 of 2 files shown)


oimiragieo:
## Round 2: both BLOCKING findings fixed (commit `240be11`)

Independent gate on `eaf61a2` was right: the first cut only fixed the non-git-repo case. Fixed both.

### BLOCKING-1 (no-op inside a git repo) — fixed

Re-ran the gate's exact repro shape against the live v1.98.8 binary (`git init` fixture, root
`.gitignore` = `*.log`, search a child `pkg/` dir with no ignore files of its own) to confirm the
mechanism before touching code:

```
=== --no-ignore-vcs (non-json) in pkg/ ===
routing_backend=RipgrepBackend  ->  pkg\keep.txt, pkg\ignored.log   (correct)

=== --json --no-ignore-vcs in pkg/ ===
routing_backend=NativeCpuBackend -> matched_file_paths: ["pkg\keep.txt"]   (BUG, confirmed)
```

Matches the gate's finding exactly. Fix: in `build_walk_builder`'s `else` branch
(`native_search.rs`), when `config.no_ignore_vcs` is set, also disable the walker's own
`git_ignore`/`git_global`/`git_exclude` knobs — not `parents`, since `--no-ignore-vcs` is
VCS-scoped only and `.ignore`/`.rgignore` parent-ascent must keep working. Applied the remedy as
specified, including folding in the `.git/info/exclude` case via `git_exclude(false)`.

### BLOCKING-2 (tests structurally incapable of catching it) — fixed

Added 2 more unit tests using a git-repo topology (`.git` marker dir + root `.gitignore` + search
a child dir with no ignore files of its own — the "best variant" the gate suggested, isolating the
native git path from `add_ignore`):
- `build_walk_builder_honors_root_gitignore_inside_git_repo_via_native_git_path` — non-regression,
  proves the native git path is live independent of `add_ignore`.
- `build_walk_builder_no_ignore_vcs_reincludes_gitignore_matched_file_inside_git_repo` — the actual
  BLOCKING-1 regression lock (RED-before-fix stated structurally, same cargo-forbidden caveat as
  round 1; the live-binary table above is the execution-verified RED for this exact topology).

The original 4 non-git tests are kept (still honest, still outcome-asserted, just were provably
insufficient on their own — now supplemented, not replaced).

### Status
- `rustfmt --check` clean on both files post-rebase.
- Rebased onto `origin/main` (`dd84bb1`, a release-only commit, no conflict) — now at `240be11`.
- Still draft, awaiting re-gate.

PR #780, opened 2026-07-26
fix(checkpoint): never destroy a file undo cannot revert (#297)

oimiragieo:
## The defect

`undo_checkpoint`'s commit phase is only crash-safe because every destructive step **first records the bytes it is about to destroy**. The `except Exception` revert at `checkpoint_store.py:1348` restores from that record and nothing else.

At three sites, a failed read set the record to `None` and the destruction proceeded regardless:

| site | what happens |
|---|---|
| current-tree file absent from the snapshot | read fails → `unlink()` still runs → nothing appended to `committed_removes` |
| snapshot-recorded-deleted file | same shape |
| restore target | read fails → `copy2` overwrites it → nothing appended to `committed_overwrites` |

In every case the file is destroyed and **the revert provably cannot bring it back**. That is precisely the permanent loss the round-7 rank-6 comment sitting directly above this code says was fixed:

> store `(path, prior_bytes)` — NOT just the path — so a commit-phase failure can recreate a file it already unlinked. **Previously the revert had only the path and permanently lost the content.**

The `OSError` arm reintroduced it. And it is reachable: the H2 pre-flight at `:1228-1250` probes readability of **snapshot** files only — working-tree targets are never probed.

Found by the #292 silent-loss census.

## The fix

`_bytes_or_abort_undo`: no capture, no destruction. The existing handler rolls back whatever was already applied and re-raises, so the tree stays consistent and the user is told which file to make readable.

New `CheckpointUndoUnsafeError`, deliberately distinct from `CheckpointCorruptError` — there the *snapshot* is unusable; here the snapshot is fine and the *working tree* is what blocked the undo.

**Also corrects the CLI error code.** The handler labelled every undo failure `checkpoint_not_found`. For this case that is a lie — the checkpoint was found and is good — and it sends the reader hunting a missing checkpoint instead of the file named in `detail`. New code: `undo_unsafe`. The same mislabel still covers `CheckpointCorruptError`; correcting *that* is a JSON-contract change and is **filed separately** rather than smuggled into a data-loss fix.

## Verification

Bidirectional. With `_bytes_or_abort_undo` returning `b""` instead of raising, both new tests fail with `DID NOT RAISE`.

The arms that make them mean something:

- **Control**: an unpatched undo genuinely *does* delete the file, asserted before the treatment runs. Without it, an undo that failed for any unrelated reason would satisfy the treatment assertions.
- **Scoped fixture**: the `read_bytes` fake fails for exactly one filename. A blanket raiser would break the snapshot machinery and abort for the wrong reason — a test that passes without proving anything.
- **The second test exercises the REVERT, not just the abort.** Removals iterate `reverse=True`, so `extra_z.py` is unlinked *and recorded* before `extra_a.py` fails; the test asserts `extra_z.py` came back with its real content. The error message claims "changes already applied have been rolled back" — this is what makes that claim testable rather than decorative.

**86 pass** across all four checkpoint suites + the census. ruff clean.

## Census ratchet

`checkpoint_store.py` 10 → 7. One of the remaining 7 is a documented **false positive**: the pre-flight probe accumulates into `missing` and then raises `CheckpointCorruptError`. That is disclosure, not loss — the detector counts `append` as an accumulator and cannot see that the collection feeds a `raise`. Recorded at the ratchet so it isn't re-litigated.


oimiragieo:
## Added: the detector, not another exemption

Auditing `main.py` for the same campaign turned up a **third** false positive of the same kind, so per this repo's own rule — *model the class, don't enumerate the cases* — I fixed the detector rather than hand-annotating a third exemption. A census that cries wolf gets its real findings waved through, which defeats the point of having one.

Two shapes the handler-local check could not see:

1. **TERMINATES** — `-f`/`--file` pattern-file reads already fail loud with exit 2 per the Backend Fail-Closed Contract, but via `_exit_search_error(...)` rather than an inline `raise`. The AST shows a call; the detector read it as silence. Nothing downstream can receive a truncated result from a branch that exits.
2. **ACCUMULATE-THEN-RAISE** — this PR's own documented false positive. The undo pre-flight appends each unreadable blob to `missing`, then raises `CheckpointCorruptError` on it once the loop ends. Disclosure built out of an accumulator; handler-local inspection can't tell it from loss. **This replaces the hand-written exemption I had added in the ratchet comment.**

### Why this is on THIS branch

It could have been its own PR, but a separate branch off `main` would have set `checkpoint_store` to 9 while this branch sets it to 7 — the two would fight over the same ratchet line depending on merge order. Branching off *this* branch instead is the documented squash hazard (a stacked PR lands half its parent). Same finding, same file, one commit.

### The arm that matters

Teaching a detector to forgive is exactly how it starts under-reporting, so the **near miss** is tested too:

```python
# raises on `kept`, while `missing` is collected and thrown away -> still counts
if not kept:
    raise ValueError("empty")
```

If that ever stops counting, the raise-name check has widened into "any raise anywhere excuses the whole function."

### Evidence it removed false positives rather than blinding the detector

Both moves were **predicted from reading the two sites**, then confirmed by re-running the census: `main.py` 18 → 17, `checkpoint_store` 7 → 6, and **nothing else shifted**. A detector that had simply gone slack would have dropped counts across unrelated files.

3 pass, ruff clean.

PR #786, opened 2026-07-26
fix(rust): cycle guard for pub-use re-export chains in _rust_resolve_use_binding

oimiragieo:
`_rust_resolve_use_binding` follows `use` re-export chains by recursing on the nested binding, with **no visited-set and no depth cap**. Rust re-exports are a GRAPH, not a tree: two modules that re-export the same name through each other close a loop and the recursion runs until the interpreter's stack dies.

### Receipt

`tg refs . <symbol> --json` on a 594-file Rust workspace exited 1 with `RecursionError: maximum recursion depth exceeded` and produced **no stdout at all** — so this took out *every* refs query on that repo, not just the cyclic symbol.

Measured before/after on the same command, same repo:

| | exit | stdout |
|---|---|---|
| unpatched | 1 | 0 bytes (`RecursionError`) |
| patched | 0 | 7209 bytes of valid JSON |

### The fix

A visited-set keyed on `(resolved importer path, imported name)`. The same file reached again for the **same** name is a cycle; the same file for a **different** name is legitimate work and must not be pruned.

### Behaviour-preserving where no cycle exists

Byte-identical output on three scopes that already worked (`hydron-paths::display_path`, `hydron-tools-agentic::uri_to_path`, `hydron-cli::main`), compared with the session daemon off so both arms took the same code path.

### Tests

Two, both bidirectionally verified against the same file with the guard reverted:

- `test_cyclic_rust_pub_use_reexport_terminates` — fails with `RecursionError` without the guard. **Reproducing it requires the searched symbol to DIFFER from the re-exported name**: a matching name hits the early return above the recursive call, which is why my first fixture passed in both arms and proved nothing.
- `test_cycle_guard_does_not_prune_a_legitimate_reexport_chain` — the control. A guard that returned early for everything would also "terminate", so this pins the other direction.

`tests/unit`: 348 passed. The one pre-existing error (`test_ast_backend.py`, `fixture 'mocker' not found` — no `pytest-mock` in my venv) reproduces identically on pristine `main`.

Draft, so CI is the arbiter — I have not merged.

### DIFF
--- a/src/tensor_grep/cli/repo_map.py
+++ b/src/tensor_grep/cli/repo_map.py
@@ -3406,7 +3406,31 @@ def _rust_resolve_use_binding(
     binding: dict[str, Any],
     symbol: str,
     repo_root: Path | str | None = None,
+    _seen: frozenset[tuple[str, str]] | None = None,
 ) -> dict[str, Any] | None:
+    # Cycle guard. This function follows `use` re-export chains by recursing on
+    # the nested binding, and Rust re-exports are a GRAPH, not a tree: a crate
+    # root that does `pub use rules::x::{...}` while a submodule re-exports back
+    # toward the root closes a loop, and the recursion never terminates.
+    #
+    # Receipt: `tg refs . <symbol> --json` on a 594-file Rust workspace
+    # (claude-code-hydron) died with `RecursionError: maximum recursion depth
+    # exceeded`, ~1000 frames of this function, rc=1 and no stdout -- so it took
+    # out EVERY refs query on that repo, not just the cyclic symbol.
+    #
+    # It DOES reproduce on a two-module fixture, but only when the searched
+    # symbol differs from the re-exported name: a matching name hits the early
+    # return above the recursive call. See
+    # test_cyclic_rust_pub_use_reexport_terminates.
+    #
+    # The key is (resolved importer path, imported name): the same file reached
+    # again for the same name is a cycle, while the same file for a DIFFERENT
+    # name is legitimate work and must not be pruned.
+    key = (str(importer_path).casefold(), str(binding.get("imported", "")).casefold())
+    seen = _seen or frozenset()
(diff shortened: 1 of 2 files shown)


oimiragieo:
## Independent gate (steward session) — PASS on all three checks

This PR was authored by a different agent, so it got an independent review rather than a self-gate. Recorded here so the verdict is not trapped in one session.

**1. Design — visited-set, not a depth cap. PASS.**
The guard keys on `(importer path, imported name)`. That distinction is the whole review: a bare depth cap would have traded a crash for a *silently truncated* answer on a legitimate deep re-export chain — the same trust defect wearing a different face, and squarely the thing task #292 exists to stop. A visited-set cannot do that, because the same file reached for a *different* name is never pruned. The inline comment already says this; it matches the code.

**2. Bidirectional oracle — control arm FAILS as required. PASS.**
Not taken on the PR's word. Ran the fixture against the unpatched tree:

```
guard present in this build? False        <- verified structurally, via co_varnames
RESULT: RecursionError  -> the bug is REAL
```

The premise was checked first (`_seen` genuinely absent) so this could not be a fixture that never applied. The docstring's warning is also correct and worth keeping: searching for `Shared` hits the early return above the recursive call, so that variant would pass with the guard removed and prove nothing. Searching an unrelated symbol is what makes the test discriminate.

**3. Not stacked on another open PR. PASS.**
Base `main`, exactly one commit (`4c54b93`). No squash-lands-half-the-work risk.

**Severity note:** this is the sharpest form of the #292 class seen so far — not an incomplete answer but *no* answer. `tg refs --json` exited 1 with a stack trace and zero stdout, taking out every refs query on that repo, not just the cyclic symbol.

**Status:** CI 0 fail / 4 pending at the time of writing. Un-drafting is gated only on those four going green; the review itself is done. Tracked as task #301, queued behind the #775..#785 drain.

PR #792, opened 2026-07-26
fix(readiness): make the exit-code gate three-state aware (#276 slice C0)

oimiragieo:
Slice **C0** of #276 — and it **gates** slice C.

## Why this comes first

An adversarial review of the #276 plan returned **HOLD** on the exact open question the plan had flagged: *is exit 2 safe?* It is not. The blast radius is **six consumers**, and I verified the three sharpest against the tree myself:

| Consumer | Site | What breaks |
|---|---|---|
| **tg spawns itself** | `rust_core/src/crossover.rs:354` — `bail!` on `!status.success()`, child from `env::current_exe()` | an exit-2 child kills `tg calibrate`, with a **misattributed** cause |
| **`tg dogfood`** (shipped command) | `scripts/agent_readiness.py:348` | ~30 `tg search` invocations; `dogfood.py:429` runs these by default; also the `windows-agent-readiness` CI gate |
| e2e byte-fidelity | `tests/e2e/test_native_json_byte_fidelity.py:100` | asserts `returncode == 0`, CI-gated via `TG_REQUIRE_RG_PARITY=1` |
| benchmarks | `run_gpu_native_benchmarks.py:371`, `run_rg_parity_benchmarks.py:62` | report `FAIL`, not partial |
| MCP | `mcp_server.py:1922` | total loss of results |

Shipping the producer first would have turned an honest partial into a broken `tg calibrate`, a failing `tg dogfood`, and a red matrix. Consumers move first; this PR does the highest-traffic one.

## The design point: allow-list, not `== 2 -> continue`

Exit 2 is **overloaded**. It is what an honest incomplete scan will return once #276 lands — but it is also what a regex syntax error returns. ripgrep's own docs say 2 is *"true for both catastrophic errors ... and soft errors."* A tolerance keyed on the bare exit code would swallow every real error and turn this sweep into a check that **cannot fail**.

So exit 2 is tolerated **only when the run disclosed why** — `result_incomplete` or `incomplete_reason_class` must be present on either stream. No marker, no tolerance. This deliberately mirrors the existing #121 carve-out immediately below it, which is scoped to one label + one precondition + one stderr substring rather than to the code.

## Verification — the red arm was proved, not assumed

```
weakened `_disclosed_incomplete` to `return True`
  -> FAILED on the control: assert not disclosed("", "regex parse error: unclosed group")
restored
  -> passed
```

42/42 in `test_agent_readiness_script.py`. `ruff format --preview` + `ruff check` clean.

Remaining C0 sites (benchmarks, the e2e assertion, `crossover.rs`, MCP) follow in sibling PRs before slice C lands.

### DIFF
--- a/scripts/agent_readiness.py
+++ b/scripts/agent_readiness.py
@@ -282,16 +282,33 @@ def _search_flag_tokens_for_sweep(command: list[str]) -> set[str]:
 # into a check that cannot fail, which is the failure mode this repo names Form 1.
 #
 # So: tolerate 2 ONLY when the run explained itself. The marker must be PRESENT. No marker,
-# no tolerance. That mirrors the existing #121 carve-out below, which is likewise scoped to
-# one label + one precondition + one stderr substring rather than to the bare exit code.
-_INCOMPLETENESS_MARKERS = ("result_incomplete", "incomplete_reason_class")
+# no tolerance.
+#
+# HONEST DIFFERENCE from the #121 carve-out below: that one is scoped to a single label AND
+# a precondition (rg absent) AND a stderr substring. This guard has only the substring --
+# it applies to all 28 labels with no precondition. That is a deliberately wider allow-list,
+# so the marker set is the ONLY thing keeping it honest; widen that set with care.
+# The JSON envelope's keys (json_fmt.py:127, :140) AND the plain-text route's stderr sentinel.
+# BOTH are required, and an independent audit caught why: every one of the 28 sweep cases is
+# plain-text (`--no-json` is even in the inverse-flag set), and NO tg route writes either JSON key
+# to stderr -- so a JSON-key-only check has a treatment arm that can never fire on the routes this
+# consumer actually runs. The plain-text disclosure is `tg: rg exited 2, keeping partial results:
+# {reason}` (ripgrep_backend.py:143, :324, :443, and the timeout variant at :187), so the sentinel
+# below is the phrase that route really emits.
+_INCOMPLETENESS_MARKERS = (
+    "result_incomplete",
+    "incomplete_reason_class",
+    "keeping partial results",
+)
 
 
 def _disclosed_incomplete(stdout: str, stderr: str) -> bool:
     """True when an exit-2 run disclosed that its scan was incomplete.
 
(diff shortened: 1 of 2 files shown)


oimiragieo:
## Independent audit: SHIP_WITH_CHANGES → both blocking findings closed

An independent Opus gate audited this against the slice-C0 plan. Both blocking findings were **real**, and I verified each against the tree rather than accepting them on report.

**Blocking 1 — the treatment arm was unreachable.** All 28 sweep cases are plain-text (`--no-json` is in the inverse-flag set), and **no tg route writes either JSON key to stderr**. A JSON-key-only check could therefore never fire on the routes this consumer runs — once slice C lands, an honest incomplete run would exit 2 carrying no marker and the sweep would *still* fail. My own test encoded the bug: `assert not disclosed("", "tg: /srv/locked: Permission denied")` is exactly what the plain-text route emits.

Fixed by adding the sentinel that route actually emits — `keeping partial results` (`ripgrep_backend.py:143`, `:324`, `:443`, timeout variant `:187`). Still an allow-list: a regex syntax error never emits that phrase.

**Blocking 2 — the test did not constrain the change.** Deleting the 5-line call-site guard left the suite **42/42 green**. Fixed with an integration pair on the existing `_make_flag_sweep_fake_run` pattern, driving the real `validate_public_search_advertised_flag_sweep`, with `_ripgrep_available` forced True so the #121 carve-out can't make it pass for the wrong reason.

Re-ran the audit's own mutation as the proof:
```
delete the call-site guard
  -> FAILED test_flag_sweep_tolerates_a_disclosed_incomplete_exit_two
restore
  -> 44 passed
```

Also corrected two comments the audit flagged as overstating — the docstring implied naming a path was a disclosure (it isn't), and the code claimed to "mirror #121" when #121 has label + precondition + substring while this guard has only the substring. That asymmetry is now stated plainly.

Non-blocking findings not taken here, with reasons: `incomplete_reason` omitted from the marker tuple is harmless (the boolean key is always co-emitted); comment/call-site distance is a style call; and the two skipped sites (`:145`, `:1077`) were independently confirmed sound for slice C's scope — both are tracked in **#313** with the audit's reasoning so the next author doesn't re-derive it.

PR #795, opened 2026-07-26
feat(native): carry a walk-error count in SearchStats (#276 slice A)

oimiragieo:
**Draft on purpose** — I cannot compile Rust on this box (CPU-SAFE), so CI is the oracle. Per the repo's own rule, a draft never un-drafts without an independent gate *and* a green matrix.

C0 is 6/6 as of #792/#793, so the producer is unblocked. This is **slice A: plumbing only, deliberately inert.** Nothing reads `walk_errors` yet, so output is byte-identical. Slice B puts it in the envelope; slice C wires the exit code.

## Two decisions worth recording

**A count, not a path list.** The walk runs in `build_parallel()` across threads. An unbounded per-path `Vec` behind a mutex would be both a contention point on the hot walker *and* a DoS surface — a tree with 50k unreadable entries produces a 50k-entry payload the agent then has to parse. A consumer only needs to know **that** the answer is incomplete.

**The merge line is the whole point.** `merge_search_stats` aggregates per-worker stats. Omitting `target.walk_errors += source.walk_errors` would have every worker count its own errors while the aggregate reports **zero** — an envelope claiming a complete result on an incomplete walk, which is exactly the defect #276 exists to fix. The comment says so at the site so it can't be dropped in a future edit.

## Verification

CI compiles. But the one break a new field actually has, I checked **mechanically** rather than from recollection:

```
grep -rn "SearchStats\s*{" rust_core/src/   ->  no struct literals
```

Every hit is `GpuNativeSearchStats` (a different type); every `SearchStats` construction is `SearchStats::default()`, so a new `usize` is Default-filled at 0. `rustfmt --check` exit 0.

Plan: `docs/plans/2026-07-26-276-native-json-incompleteness.md` (#794).

### DIFF
--- a/rust_core/src/main.rs
+++ b/rust_core/src/main.rs
@@ -8359,10 +8359,17 @@ fn run_native_search_with_optional_rg_fallback(
             // 1, which reads as an authoritative "no matches exist" -- the exact lie #276 exists
             // to stop.
             //
-            // Safe to emit only because slice C0 landed first: all six exit-code consumers are
-            // three-state aware as of #792/#793 (agent_readiness x3, both benchmark harnesses,
-            // the byte-fidelity e2e, mcp_server), and crossover.rs was verified unreachable.
-            // Shipping this before them would have broken `tg calibrate` and `tg dogfood`.
+            // 🔴 MERGE PRECONDITION -- NOT satisfied history. PRs #792 and #793 make the six
+            // exit-code consumers three-state aware (agent_readiness x3, both benchmark
+            // harnesses, the byte-fidelity e2e, mcp_server; crossover.rs verified unreachable).
+            // AS OF THIS COMMIT BOTH ARE STILL OPEN, and `origin/main` still has
+            // `agent_readiness.py` rejecting any exit not in {0,1}. THIS BRANCH MUST NOT MERGE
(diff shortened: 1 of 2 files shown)


oimiragieo:
Un-drafted: the merge precondition is now satisfied.

This PR was held because it makes the native binary exit 2 on walk errors, which had to land **after** #793. #793 merged at `2026-07-26T21:51:59Z`, so that ordering constraint is discharged.

State: **48 checks green**. Next in the one-per-publish drain behind #785/#800/#801, gated on release run `30221952105` finishing (currently queued on macOS runners).

This is the producer half of #276 — `SearchStats::walk_errors` plus the three additively-conditional envelope fields. Every remaining #276 slice (#314, #315, #316, #317, #318, #320, #321) is blocked on it, because the fields they extend do not exist on `main` until this lands.

Note for whoever picks up those slices: the plan at `docs/plans/2026-07-26-276-json-incompleteness-remaining-slices.md` (#803) was revised after review falsified two of its tasks — #319's defect is unreachable (ships as defense-in-depth in #805), and #321 is a **live** defect that an earlier draft wrongly called mis-filed. Read the post-mortems in Tasks 1 and 2 before building.

PR #797, opened 2026-07-26
test(determinism): make output stability a named, CI-gated invariant (#311)

oimiragieo:
Closes #311. A golden test runs the command ONCE against a stored expectation — it catches drift FROM the golden, never run-to-run INSTABILITY. A field emitted in set-iteration order passes whenever it happens to match and reads as a FLAKE when it does not, which is why it gets re-run instead of diagnosed.

7 machine-facing surfaces x 3 runs in SEPARATE PROCESSES under PYTHONHASHSEED 0/1/2.

- Separate processes because an in-process repeat inherits lru_cache/globals/warm session state and would be trivially identical — a check that cannot fail.
- Hash seeds VARIED deliberately: set/dict order changes between seeds, so this reliably detects the most common nondeterminism source in Python instead of coin-flipping it.
- The volatile set is an allow-list split by ALWAYS vs CONDITIONAL. Normalising away whatever differs is how such a gate becomes decoration; `generated_at` has its liveness asserted, `daemon_response_cache` is declared conditional and documented as not liveness-checkable rather than loosening an assertion until it passed.

**MEASURED:** 4/7 surfaces byte-identical with NO normalisation. The other 3 differ only in provenance/runtime metadata. Results were already deterministic — this stops that regressing silently.

**PROVEN:** two injections went RED naming the JSON path. Three earlier injections were INVALID (dict-not-set, markdown-only path, sidecar module never routed to) — recorded in the docstring with the rule they share.

No new CI job: tests/unit is already covered by test-python.


oimiragieo:
CI RED decoded — and the finding is the **opposite** of what the failure said.

`codemap-json is NOT deterministic -- 19 field(s) moved` was **this gate's own bug**, not a product defect. All 19 "moved" fields were the per-run scratch directory: `str(sandbox)` never appeared in the emitted text, so all three replaces missed, the path survived into the comparison, and two runs with different temp dirs duly differed. Linux/macOS passed, so it read as a genuine platform-specific defect.

Two fixes in `d781fc2`, the second the important one:

1. Normalise the **resolved** spelling as well as the handed one — the tool resolves the path it is given, and where the two differ (8.3-shortened or junctioned TEMP, casing, UNC prefix) replacing only the handed form matches nothing.
2. **Assert normalisation actually fired.** A gate that misreports its own gap as a product finding is worse than no gate — it is confidently wrong in the product's voice, and it cost a full CI cycle plus a wrong root cause. The marker is the scratch prefix, now a single constant used both to create the sandbox and to detect it.

Change 1 is a *hypothesis* about which spelling CI emits — it does not reproduce on my Windows box, where the two spellings are identical. Change 2 is what makes shipping that acceptable: if the hypothesis is wrong, the next run prints `NORMALISATION GAP` **with the excerpt containing the third spelling**, instead of accusing the product again.

Verified bidirectionally in one process: as written 8 passed; with both spellings made non-matching (reproducing the CI condition) it goes red with `NORMALISATION GAP`, not "N field(s) moved".

PR #803, opened 2026-07-26
docs(plan): implementation plan for #276's remaining seven slices

oimiragieo:
**Draft on purpose** — a planning artifact, not a code change, and I am at the WIP cap.

## Read this first: the review falsified two of my own tasks

I published a first draft claiming two findings. Both were wrong, in opposite directions, and the corrections are the most useful thing in this PR.

**#319 — I claimed a defect that cannot occur.** I said the `Drop` guard omits `binary_match_files` and silently drops a worker. It does omit it, but `searched_files += 1` (`:536`) precedes `binary_match_files += 1` (`:547`), so `binary_match_files > 0 ⟹ searched_files > 0` and the guard can never fire. My proposed test reached that state only by assigning the field directly — an oracle with no discriminating power, which is the exact fault I flag in other people's tests. The refactor survives as defense-in-depth, honestly labelled.

**#321 — I claimed a real defect was mis-filed.** It is live. The error at `:1378` is `anyhow::Error` (`search_path` returns `anyhow::Result`, `:509`); `ignore::Error` never reaches there (`:13` is the only `ignore` import); and `is_io()` exists on neither — my proposed fix would not have compiled. The actual bug: stdout write failures (`:546`, `:554`) pass a guard matching only typed `BrokenPipe` (`:307-311`) and get published as `incomplete_reason_class: "unreadable_path"` (`:2437`) — telling an agent a *path* was unreadable when *our own output* failed. A wrong disclosure is worse than a missing one. A live exit-code regression rides along: `main.rs:8373` exits 2 where `BROKEN_PIPE_EXIT_CODE = 1` was intended.

**On the split:** two of three lenses voted "mis-filed". Both reasoned only from my draft's type error without tracing what actually reaches the site; one traced it. I verified that chain myself before overruling the majority. Majority is not evidence.

## Other corrections

- **#320 reversed.** I proposed `HashSet<PathBuf>`. `native_search.rs:87-90` documents that exact design as rejected (mutex contention + DoS at 50k unreadable entries) and I neither cited nor rebutted it. Quantified: ~8 bytes today vs ~8.5 MB at 50k, ~170 MB at 1M. `SearchStats` derives `Serialize`, so a `HashSet` would also break byte-reproducibility and the #311 determinism gate. And my "renaming costs a major bump" premise was false — `CONTRACTS.md:171` binds *existing* fields, and a repo-wide grep for `incomplete_paths_count` returns only this plan. **The rename is free.**
- **#314's record shape is determined, not chosen.** It must carry `version`/`routing_backend` (`test_harness_adoption.py:105-109`) and must omit `file`/`path` (`test_native_renderer_file_set_invariant.py:143-164`, or the search root is injected into the file set). `tests/helpers/rg_parity.py:525` hard-KeyErrors on any row lacking `file` — one unavoidable consumer edit. `--format rg --ndjson` never reaches rg (`main.py:5309`) and rg's summary event has no incompleteness field, so there is no contract to match.
- **Four wrong anchors** corrected: `collect_walked_files` `:1621`→`:1709`, `search_ndjson` `:619`→`:629`, `merge_search_stats` `:1413`→`:1404`, Task 5's range.
- **A whole second envelope** was uncovered: `SearchResultJson` at `main.rs:12987`, six live call sites, none of the three fields, and `main.rs` appeared in no task.
- **Two unowned obligations** are now tasks: exit-code parity (the exit-2 rule is reachable from one route only) and the MCP contract bump (`mcp_server.py:120` has zero occurrences of `incomplete_reason_class`).

## Dependency reality

Most slices sequence behind #795 → #793 → the drain gate, because `walk_errors` and the envelope fields do not exist on `main`. **Exception the review found:** the `is_empty()` refactor targets code on `main` today (`origin/main:native_search.rs:686-692`) — it is the refactor #795 should rebase *onto*, not a dependant.


oimiragieo:
Closing as **superseded** — every slice this plan schedules has shipped, so merging it would add a document describing pending work that is already done.

Verified against `origin/main` per slice, not inferred from the task list:

| planned slice | shipped as |
|---|---|
| ndjson / `main.rs` envelopes (314) | #811 |
| `collect_walked_files` count channel (315) | #808 |
| `gpu_native` twin (316) | #823 |
| multi-pattern JSON seam (317) | #818 |
| `CONTRACTS.md` contradiction (318) | #815 |
| `SearchStats::is_empty()` (319) | #805 |
| `incomplete_paths_count` counts EVENTS (320) | #820 |
| output-write failure ≠ unreadable path (321) | #821 |

0 of 8 missing.

The narrative record lives in `docs/BACKLOG.md` instead (#828), which describes #276 as closed across nine PRs and links each one. A merged plan-for-completed-work is the same defect as the stale "IN FLIGHT" block #828 deletes: it tells a reader work is pending that already shipped.

Reopen if a slice turns out to be incomplete — the branch is untouched.

PR #822, opened 2026-07-27
fix(cli): lead, re-label, and correct the knob on every truncation disclosure the symbol commands emit

oimiragieo:
Closes task #329.

## The defect

`tg defs/refs/callers/impact` and `tg blast-radius` disclose a truncated scan by
appending `warning: INCOMPLETE RESULT: ...` **after** the payload. A trailing
marker is the easiest thing to emit and the most ignored: the consumer reads the
prefix as the document and the last line as a footnote, so a caller-set truncated
at a file cap gets trusted as exhaustive — exactly the wrong-refactor risk the
exit-2 gate exists to prevent. The exit code and the JSON `result_incomplete`
field were already right; only the text-path *position* was wrong.

## The fix

`_completeness_caveat_lines(caveat, *, is_truncation)` returns
`(leading_banner, trailing_note)` and is the single definition of the ordering:

| caveat | position | why |
|---|---|---|
| truncation (`warning:`) | **leads** the payload | must be read *before* the data it qualifies |
| zero-callers advisory (`note:`) | **trails** the payload | that result is COMPLETE; the note only warns against over-reading it |

Applied at **both** text emitters — `_emit_symbol_command_result` and the
`blast-radius` counts block. Defining it once is deliberate: the twin site is how
this class of defect recurs (an agent reading `callers=3` first has already formed
the answer by the time a trailing line lands).

JSON output is untouched — `caveat` is a field there and field order carries no
reading bias. Pinned by a test so a future change can't leak a banner line onto
stdout and break `json.loads`.

## Verification (bidirectional, control arm run)

New `tests/unit/test_leading_truncation_banner.py`, 7 tests. With the split
reverted to trailing (mutation asserted applied, not assumed):

```
FAILED test_truncation_warning_leads_the_blast_radius_counts
FAILED test_truncation_warning_leads_the_symbol_payload
FAILED test_completeness_caveat_lines_puts_truncation_first_and_advice_last
3 failed, 4 passed
```

The 4 that keep passing are the control/advisory/JSON arms — so the suite
discriminates rather than observing "a marker appeared somewhere in stdout"
(oracle Form 7: `assert "warning:" in out` passes identically in both arms).
Every ordering assertion is preceded by a premise assert that the payload line
was actually rendered, so an inert `emit_text` can't make `index()` comparisons
vacuously true.

Neighbouring suites green: `test_main_cli_contracts` +
`test_answer_first_symbol_payloads` + `test_deadline_blind_truncation_gates` +
`test_cap_fix_chokepoint` → 102 passed. `ruff check` + `ruff format --check
--preview` clean.

## Docs

AGENTS.md gains **"A Disclosure Must Precede The Data It Qualifies"** — the
reusable law plus the two consequences (test position, not presence; define the
ordering once and share it across emitters).


oimiragieo:
@codex review

oimiragieo:
## Independent Opus gate: SHIP-WITH-FIXES — both blockers fixed in d52bf86

Transcribing the gate's **non-blocking** findings here, because a subagent gate's result is
ephemeral: acting only on its blockers silently drops everything it classified non-blocking.

### Fixed in this PR (d52bf86)
- **B1** `_completeness_caveat_lines`' docstring said "the two text emitters" after this PR wired a third.
- **B2** `AGENTS.md` claimed the helper serves "every text emitter" — false of most of the CLI.
- **N4a** The comment credited the wrong mechanism for `%%{` safety (the space *after* `%%`, not the indentation).
- **N3** Banner flattened against newline injection from payload-derived values; pinned by a discriminating test.
- **N6** Added a command-level `--mermaid` test (all prior mermaid tests called the renderer only).
- **N5** Test docstring no longer claims a red-green that cannot be reproduced (reverting deletes the imported helper → the file ERRORS, which is not a red arm).

### NOT fixed here — filed, with the gate's evidence

**N1 — the `--deadline` arm gets exit 2 with ZERO disclosure.** Pre-existing and identical in both
arms, so not a regression, but it is the largest ABSENT case in the class this PR names.
`_scan_truncation_warning` reads only `scan_limit`/`caller_scan_limit`/`output_limit`, while
`_scan_incomplete` *also* returns True on `payload["partial"]`; `repo_map.py` stamps
`partial=True` + `deadline_limit` on a deadline cutoff **without** calling `_mark_result_incomplete`.
So `_annotate_result_completeness` returns `(None, False)` and every branch stays silent:

```
ARM A (CONTROL, scan_limit cap)      exit 2 -> "warning: INCOMPLETE RESULT: the scan stopped at a 512-file cap..."
ARM B (partial=True + deadline_limit) exit 2 -> "Blast radius for x in .\ndefinitions=1 callers=1 ..."   (silent)
ARM B2 (same, --mermaid)              exit 2 -> graph TD / target / n0 / n0 -->|L9| target                (no %% line)
```

**N7 — uncrossed siblings in the same family.** `blast-radius-plan` and `blast-radius-render` both
exit 2 with the identical counts-then-nothing shape. The gate confirms there is **no missed twin
inside `blast-radius` itself** — the renderer has exactly two production call sites — so the
enumeration for *this command* is complete; the gap is one command over.

**Scope note (N4b).** A `%%` comment is source-level metadata. This fix serves the
agent-reading-the-source audience; a viewer of a *rendered* diagram still sees no disclosure either
way. Filed separately — the disclosure should also be a visible graph node.

All three are covered by the text-mode-disclosure class-fix task, which the census independently
sized at ~12 affected surfaces (4 HIGH). They are deliberately not folded into a rendering fix.

PR #830, opened 2026-07-27
docs(contracts): stop excluding two routes from the allow-list that both landed

oimiragieo:
docs(contracts): stop excluding two routes from the allow-list that both landed

`docs/CONTRACTS.md`'s "WHAT REMAINS TRUE" bullet still told agents that "two
routes still discard the count and are tracked, so they belong in neither list
until they land: the multi-pattern `-e`/`-f` JSON route (task 317) and the
CUDA-gated `gpu_native.rs` twin (task 316)". Both landed -- #811 and #823 -- so a
load-bearing contract was instructing consumers to distrust two surfaces that had
been fixed. Wrong in the UNSAFE direction is the sibling failure this same bullet
was rewritten for once already.

## Why it rotted, which is the reusable part

`test_contracts_native_json_incompleteness_claims_match_the_rust_source` pins
THREE retracted phrases in this paragraph and not the sentence sitting beside
them. That is plan Task 10's finding reproducing itself inside the very bullet it
was written about: pinning some claims of a multi-claim paragraph leaves the rest
free to rot, and the unpinned ones are exactly where drift lands.

## The exclusion was never actionable

Recorded rather than quietly deleted, because the reason generalises. Both
envelopes stamp `routing_backend` from `decision.routing_backend()`
(`routing.rs`), so the "excluded" multi-pattern route reports the same
`"NativeCpuBackend"` string as the included single-pattern one. A consumer
following this bullet's own advice -- allow-list on `routing_backend` -- had no
field on the wire that could tell them apart. An exclusion has to be expressible
in a field the consumer actually receives; if it is not, it is not a caveat, it
is a footnote that reads as one.

## Verification (bidirectional, control arm run)

New `test_contracts_does_not_still_exclude_the_two_routes_that_have_landed`,
pinned against the RUST SOURCE in both directions like its sibling -- remove the
emission and the premise fails pointing at the Rust; let the exclusion return and
the claim fails pointing at the doc. Cited by SYMBOL (`struct SearchResultJson`,
`struct GpuNativeSearchResultJson`) rather than line number, per the #764 receipt
that prose line numbers into a churning file rot by construction.

  treatment                                  20 passed
  control (doc reverted via `git apply -R`)   1 failed, 19 passed

Exactly one test moves, and it is the new one -- so the pin discriminates rather
than riding on a neighbour.

`ruff check` + `ruff format --preview --check` clean.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>


oimiragieo:
@codex review

